PR #5, opened 2020-06-25
Listing, MapLocation, User objects

LindseyWeiskopf:
Summary: Created the Listing, MapLocation, and user object classes

Listing class: This class contains the various attributes that will be needed for each listing on the webapp: name, location, image, ratings, and website. Right now, the class just contains accessor methods for all of the instance variables.

MapLocation class: This class essentially contains a coordinate (lat and long) to be able to track users and listings.

User class: As of right now (Signed out user), this class contains an overloaded constructor so that a user can be initialized with a MapLocation (result of Geolocation API/coordinate) or a zip code.

Thanks!

### DIFF
--- a/src/main/java/com/google/sps/data/Listing.java
+++ b/src/main/java/com/google/sps/data/Listing.java
@@ -1,0 +1,48 @@
+package com.google.sps.data;
+
+/** Class for Listing object with the specified attributes returned from the places API **/
+public class Listing{
+    private String name;
+    private MapLocation mapLocation;
+    private int rating;
+    private String image;
+    private String url;
+
+  /**
+  * Creates a new Listing
+  * @param name Name of business
+  * @param mapLocation Location of business (lat/long coordinate)
+  * @param rating Numerical rating of business (1-5)
+  * @param image Image
+  * @param url Url to business' site
+  * @return Listing object
+  **/
+
+  public Listing(String name, MapLocation mapLocation, int rating, String image, String url) {
+    this.name = name;
+    this.mapLocation = mapLocation;
+    this.rating = rating;
+    this.image = image;
+    this.url = url;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public MapLocation getMapLocation() {
+    return mapLocation;
+  }
+
+  public int getRating() {
+    return rating;
+  }
+
+  public String getImage() {
+    return image;
+  } 
+
+  public String getUrl() {
+    return url;
+  }
+}
--- a/src/main/java/com/google/sps/data/Listing.java
+++ b/src/main/java/com/google/sps/data/Listing.java
@@ -1,12 +1,13 @@
 package com.google.sps.data;
 
 /** Class for Listing object with the specified attributes returned from the places API **/
-public class Listing{
+public class Listing {
     private String name;
     private MapLocation mapLocation;
     private int rating;
     private String image;
     private String url;
+    private String businessType;
 
   /**
   * Creates a new Listing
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/data/Listing.java
```diff
@@ -0,0 +1,48 @@
...
+  * @return Listing object
+  **/
+
+  public Listing(String name, MapLocation mapLocation, int rating, String image, String url) {
+    this.name = name;
+    this.mapLocation = mapLocation;
+    this.rating = rating;
+    this.image = image;
+    this.url = url;
+  }
```
ethanimooney, line 27: I might add a section for a business description. Not sure if we can pull that from the Places API in the long run, do you?

src/main/java/com/google/sps/data/Listing.java
```diff
@@ -0,0 +1,48 @@
+package com.google.sps.data;
+
+/** Class for Listing object with the specified attributes returned from the places API **/
+public class Listing{
```
ethanimooney, line 4: Super, mega, big, important nit .... space before the bracket :)

src/main/java/com/google/sps/data/Listing.java
LindseyWeiskopf: The Places API does not return a business description, but it does return a "type" like food and establishment which signifies that the entity is a business which might be even more helpful down the road. I added a businessType parameter to account for that. I didn't see anything regarding business description on the places API, but that's definitely something we can add in the future if we find a way to get that info.

src/main/java/com/google/sps/data/Listing.java
LindseyWeiskopf: Ugh! How could I miss that!!

src/main/java/com/google/sps/data/Listing.java
ethanimooney: Works for me, thx for the explanation

src/main/java/com/google/sps/data/Listing.java
```diff
@@ -0,0 +1,55 @@
+package com.google.sps.data;
+
+/** Class for Listing object with the specified attributes returned from the places API **/
```
shaquille-hall, line 3: nit: You can remove "Class for".

src/main/java/com/google/sps/data/Listing.java
shaquille-hall: +1

src/main/java/com/google/sps/data/MapLocation.java
```diff
@@ -0,0 +1,18 @@
+package com.google.sps.data;
+
+/** Class representing a MapLocation as a coordinate **/
```
shaquille-hall, line 3: nit: You can reduce this to "MapLocation representing coordinate values"

src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,32 @@
+package com.google.sps.data;
+
+/** Class representing a user with a location for v1 **/
```
shaquille-hall, line 3: nit: You can remove the note that it's a class.

src/main/java/com/google/sps/data/Listing.java
LindseyWeiskopf: Changed!

src/main/java/com/google/sps/data/MapLocation.java
LindseyWeiskopf: Got it

PR #6, opened 2020-06-26
Create Backend structure

LindseyWeiskopf:
Summary: Create back end structure for SmallCity to separate work areas and define flow between objects.

Together we created the SmallCityService.java file and the DataServlet.java file. The goal of the SmallCityService class is to return a list of the small business listings to be displayed given a User input. This class will handle processing the listings returned from the places API and removing any large businesses (in an efficient way) from the static list that we create. 

The Data Servlet will get the location input/geolocated from the user to create a SmallCityService instance, and then return the list of businesses generated from the object.

Finally, we realized that we need to convert the zipcode to a map location in the User class, so we changed the structure to convert an inputted zipcode to a coordinate using the Places API.

### DIFF
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -1,10 +1,15 @@
 package com.google.sps.data;
 
+/** SmallCityService object representing all components of the webapp **/
 public class SmallCityService {
   
   private User user;
   private List<Listing> businessList;
-
+  
+  /** Create a new Small City Service instance
+  * @param user User of webapp with mapLocation
+  * @return List of small local businesses
+  **/
   public SmallCityService(User user) {
     this.user = user;
     getAllBusinesses();
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -4,7 +4,7 @@ package com.google.sps.data;
 public class SmallCityService {
   
   private User user;
-  private List<Listing> businessList;
+  private List<Listing> businessesList;
   
   /** Create a new Small City Service instance
   * @param user User of webapp with mapLocation
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -1,0 +1,23 @@
+package com.google.sps.data;
+
+public class SmallCityService {
+  
+  private User user;
+  private List<Listing> businessList;
+
+  public SmallCityService(User user) {
+    this.user = user;
+    getAllBusinesses();
+    eliminateBigBusinesses();
+  }
+
+  public void getAllBusinesses() {
+    // TODO: Get businesses from Place API given user location
+    
+  }
+
+  public void eliminateBigBusinesses() {
+    // TODO: Parse big business list and remove big businesses from businessList
+
+  }
+}
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -4,14 +4,14 @@ package com.google.sps.data;
 public class SmallCityService {
   
   private User user;
-  private List<Listing> businessesList;
+  private List<Listing> businesses;
   
   /** Create a new Small City Service instance
-  * @param user User of webapp with mapLocation
+  * @param mapLocation geolocation of user
   * @return List of small local businesses
   **/
-  public SmallCityService(User user) {
-    this.user = user;
+  public SmallCityService(MapLocation mapLocation) {
+    this.user = new User(mapLocation);
     getAllBusinesses();
     eliminateBigBusinesses();
   }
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/sps/data/SmallCityService.java
```diff
@@ -0,0 +1,28 @@
+package com.google.sps.data;
+
+/** SmallCityService object representing all components of the webapp **/
+public class SmallCityService {
+  
+  private User user;
+  private List<Listing> businessList;
```
shaquille-hall, line 7: nit: Would "businesses" be clearer?

src/main/java/com/google/sps/data/User.java
```diff
@@ -19,14 +18,10 @@ public User(MapLocation geolocation) {
   * @return User with zipCode
   **/
   public User(int zipCode) {
-    this.zipCode = zipCode;
+    // TODO: Convert zipcode to MapLocation
```
shaquille-hall, line 21: I think you can keep the assignment from the existing version, and create a private method to handle the conversion.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -23,27 +23,21 @@
...
-  private String convertToJsonUsingGson(ArrayList<String> temp) {
-    Gson gson = new Gson();
-    String json = gson.toJson(temp);
-    return json;
+
+    @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // TODO: Create smallCityService instance from User inputted location
+    
+    smallCityService = new SmallCityService(User user);
```
shaquille-hall, line 40: What will be the goal here?

src/main/java/com/google/sps/data/SmallCityService.java
LindseyWeiskopf: yes - just changed

src/main/java/com/google/sps/data/User.java
LindseyWeiskopf: I abstracted out the helper function, but I think that this conversion may have to take place in the javascript anyways in order to use the Geocoding API to convert from zipcode to coordinates.

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: The goal here is to take the user inputted location from the JSON on the /data page to create the SmallCityService object. Does that make sense?

src/main/java/com/google/sps/data/SmallCityService.java
```diff
@@ -0,0 +1,28 @@
+package com.google.sps.data;
+
+/** SmallCityService object representing all components of the webapp **/
+public class SmallCityService {
+  
+  private User user;
+  private List<Listing> businessesList;
```
shaquille-hall, line 7: I also mean removing the indication that it's a list from the variable. :)

src/main/java/com/google/sps/servlets/DataServlet.java
shaquille-hall: Yes. But if that's the case, is there a reason to not pass the user location directly, rather than the entire user? I'm not suggesting this is wrong, understanding this is still framework code, but I'm trying to get an understanding of the high-level flow.

src/main/java/com/google/sps/data/SmallCityService.java
LindseyWeiskopf: Oh yep that would make a lot more sense! Fixed!

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: Ah I see what you mean. I think passing in the Map Location makes a lot more sense here. We'll just have to create that mapLocation object in the servlet from the lat and long values returned through the API in javascript. Should be changed now!

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -23,27 +23,21 @@
...
-  private String convertToJsonUsingGson(ArrayList<String> temp) {
-    Gson gson = new Gson();
-    String json = gson.toJson(temp);
-    return json;
+
+    @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // TODO: Create smallCityService instance from mapLocation inputted location communicated through JSON
+    // TODO: Convert lat and long maps variable to a mapLocation object
+    smallCityService = new SmallCityService(geolocation);
```
shaquille-hall, line 40: I don't think this builds. It looks like geolocation is never defined.

LindseyWeiskopf: It's not defined yet - that was more of a placeholder because we need to write more of the javascript to know where to fetch the lat and long coordinates from. Should we take that construction out all together until we write more code?

shaquille-hall: Gotcha. Since that's the case, let's take it out, or convert it to a comment. Because this is a shared repo, and serves as the source of truth for the project, only code that works should be submitted.

LindseyWeiskopf: I just deleted it. Going to make issues for the todos now.

PR #11, opened 2020-07-01
Near complete frontend mvp

ethanimooney:
TLDR: Added most all of the front end functionality, minus JavaScript.
LINK: http://step2020-smallcity.appspot.com/

**README.md:** Made more descriptive

**index.html**: Added the main divs: main-container encompasses everything but the header, creating a 3 column grid layout. map-container will, well, contain the map. results-container is a one column, 4 row layout for the results cards and pagination buttons. filters-container is where filters will live when added.

**style.css:** Styled the page according to the mocks in go/smallcity. **Note for reviewers:** the css looks long, but there are just multiple screen breakpoints, meaning that you can simply review one breakpoint as the others are the same except for a few padding/sizing changes. 

**Note*** : Pagination buttons are of different design. I did this just to be able to get them on the page, as making them float will take more time

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,16 @@
 # SmallCity
+
 | Created by Abakir Hanna, Ethan Mooney, and Lindsey Weiskopf
 
+
 ### Objective
+
 We aim to create a web platform for users to find products from local small businesses. Our web app will increase exposure to local businesses of interest in their community. 
 
+
 ### Background
+
 The U.S Small Business Administration Office of Advocacy defines a small business as a company with fewer than 500 employees. These small businesses are an integral part of the world’s economy and are often overlooked. Making up 44% of economic activity in the US, small businesses often struggle to keep up with the growing demand for instant delivery via online shopping. Only amplified during crises like the Covid-19 pandemic, the uphill battle that these businesses face poses a real threat to their stability and, in turn, the US economy.
 
 Right now, many small local businesses get lost in the shadow of large businesses on online shopping platforms. But, as buyers continue to purchase more products from the comfort of their own homes, we want to give them an option to support the businesses in their community with ease. In a 2020 survey by Salesforce, 86% of consumers reported that they would pay more to support a small local business. Giving users the ability to filter for small local businesses on an already trusted platform like Google Shopping would allow users to shop locally from wherever they are while boosting local economies. 
+
--- a/README.md
+++ b/README.md
@@ -1,9 +1,10 @@
 # SmallCity
-This directory is where we will write all of your code!
+| Created by Abakir Hanna, Ethan Mooney, and Lindsey Weiskopf
 
-By default it contains a barebones web app. To run a local server, execute this
-command:
+### Objective
+We aim to create a web platform for users to find products from local small businesses. Our web app will increase exposure to local businesses of interest in their community. 
 
-```bash
-mvn package appengine:run
-```
+### Background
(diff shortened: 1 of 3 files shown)


src/main/webapp/style.css
```diff
@@ -8,56 +8,570 @@ body {
...
-/*
-  Map Content Container Styling
-*/
-.map-container {
-  background-color: #e5e5e5;
-  border-radius: 30px;
-}
+    header {
+      align-items: center;
+      background-color: #b3ffb3;
```
LindseyWeiskopf, line 19: If you're interested (and only if you're interested), you could set these main colors to variables in the root of the css [file](https://stackoverflow.com/questions/1875852/how-can-i-define-colors-as-variables-in-css). This would make the document a little bit more readable. I'm not sure if this is good form or not but just a thought!

src/main/webapp/index.html
```diff
@@ -8,13 +8,38 @@
...
+          <button class="results-website-button">Visit Website</button>
+        </div>
+        <div class="results-card">
+          
+        </div>
+        <div class="results-card">
+          
+        </div>
+        <div class="pagination-container">
+          <button class="pagination-buttons">Back</button>
```
LindseyWeiskopf, line 40: nit: Previous and next might make a little bit more sense then forward and back

ethanimooney: I agree, changed

src/main/webapp/style.css
ethanimooney: I think I'll leave them as-is for now but may end up changing them later once I see if I can change around the screen breakpoints. Good suggestion though, never thought of that.

src/main/webapp/index.html
```diff
@@ -8,13 +8,38 @@
...
+      </div>
     </header>
 
     <div class="main-container">
       <div class="map-container">
       </div>
       <div class="results-container">
+        <div class="results-card">
+          <div class="results-image">
+            image
```
shaquille-hall, line 23: Should this say "image"?

src/main/webapp/index.html
```diff
@@ -8,13 +8,38 @@
...
+        your zipcode: 
+      </div>
     </header>
 
     <div class="main-container">
       <div class="map-container">
       </div>
       <div class="results-container">
+        <div class="results-card">
+          <div class="results-image">
```
shaquille-hall, line 22: This won't work once you start using dynamic content. I suggest using javascript to create these page elements, and assign the correct details. Similar to what you did in your portfolio project.

src/main/webapp/style.css
shaquille-hall: This will also prevent having to repeat the styles for the same elements. @media selectors should work like inheritance, in that one only needs to redefine the changing values, not the entire suite. I think this will make like easier going forward. 

What do you mean changing around the screen breakpoints? The values for min-device-width etc?

src/main/webapp/index.html
ethanimooney: Yes that's the plan, I just had some in there as placeholders to get the feel for the page.

src/main/webapp/style.css
ethanimooney: Yeah, I'm still playing around with the widths. I realize about the inheritance now, I wasn't thinking about it right, will change.

README.md
```diff
@@ -1,9 +1,10 @@
 # SmallCity
-This directory is where we will write all of your code!
+| Created by Abakir Hanna, Ethan Mooney, and Lindsey Weiskopf
```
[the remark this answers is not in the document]
ethanimooney: Got it

src/main/webapp/index.html
ethanimooney: I just had it there as placeholder text so everyone knew what it was. Will be removed once we add in the js functionality

shaquille-hall: Gotcha. Then please prefer a comment, that should serve the same purpose, over placeholder text that will appear in the dom.

ethanimooney:
Fixed the issues, removed the filters section, pruned the css down **correctly** using @media this time, added support for smaller breakpoints.

ethanimooney:
Also added the spaces in README

PR #12, opened 2020-07-01
Map implementation prototype

LindseyWeiskopf:
Summary: Complete the location services prototype without APIs including hardcoded listings and frontend to backend connections

In this pull request, I implemented geolocation services in script.js. I used the HTML5 Geolocation API to prompt the user to be geolocated, and then it creates an XMLHttp Post Request with the latitude and longitude coordinates found. This request triggers the doPost() function in the data servlet through the query string `/data?lat=" + lat + "&lng=" + lng`. This post request is necessary to get the user's location information from the client to the server side. The doPost function gets the lat and lng parameters, and creates a mapLocation instance & a SmallCityService instance from the coordinate.

From there, the SmallCityService class generates a list of all businesses in the area--six hardcoded listings as a prototype in this iteration--from the mapLocation. In this pull request, the mapLocation is also hardcoded (as Google Pitt office) because my location would not be accurate to the business results.

The business list from the smallCityService instance is then sent to the front end through json on the /data page. This transfers the businesses from the back end to the front end.

In Listing.java, I changed the ratings from integers to doubles.

Thanks!

### DIFF
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -14,11 +14,11 @@ public class SmallCityService {
   **/
   public SmallCityService(MapLocation mapLocation) {
     this.user = new User(mapLocation);
-    getAllBusinesses();
+    findAllBusinesses();
     eliminateBigBusinesses();
   }
 
-  public void getAllBusinesses() {
+  public void findAllBusinesses() {
     // TODO: Get businesses from Place API given user location
     businesses = new LinkedList<Listing>();
     businesses.add(new Listing("LA Fitness", new MapLocation(40.457091, -79.915331), 3.9, null, "https://www.lafitness.com/Pages/Default.aspx"));
@@ -34,7 +34,7 @@ public class SmallCityService {
 
   }
 
-  public List<Listing> businesses() {
+  public List<Listing> getBusinesses() {
     return businesses;
   }
 }
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -1,5 +1,6 @@
 package com.google.sps.data;
 import java.util.List;
+import java.util.LinkedList;
 
 /** SmallCityService object representing all components of the webapp **/
 public class SmallCityService {
@@ -19,11 +20,21 @@ public class SmallCityService {
 
   public void getAllBusinesses() {
     // TODO: Get businesses from Place API given user location
-    
+    businesses = new LinkedList<Listing>();
+    businesses.add(new Listing("LA Fitness", new MapLocation(40.457091, -79.915331), 3.9, null, "https://www.lafitness.com/Pages/Default.aspx"));
+    businesses.add(new Listing("west elm", new MapLocation(40.456279, -79.915015), 3.6, null, "https://www.westelm.com"));
+    businesses.add(new Listing("McDonalds", new MapLocation(40.459450, -79.918479), 2.6, null, "https://www.mcdonalds.com"));
+    businesses.add(new Listing("East End Brewing Company", new MapLocation(40.459391, -79.911782), 4.7, "https://www.google.com/url?sa=i&url=https%3A%2F%2Fwww.eastendbrewing.com%2F&psig=AOvVaw0kX_SAlxhA09EN3cKpt5ik&ust=1593613487774000&source=images&cd=vfe&ved=0CAIQjRxqFwoTCLDA6oDfqeoCFQAAAAAdAAAAABAD", "http://www.eastendbrewing.com/"));
+    businesses.add(new Listing("The Shiny Bean Coffee & Tea", new MapLocation(40.496328, -79.944862), 4.9, "https://goo.gl/maps/AYH2QCL7pkoMBxHA8", "https://theshinybean.com/"));
+    businesses.add(new Listing("Weisshouse", new MapLocation(40.456684, -79.925499), 4.3, "https://goo.gl/maps/7tuXn7QF2hh7ioGYA", "https://www.weisshouse.com/"));
   }
 
   public void eliminateBigBusinesses() {
     // TODO: Parse big business list and remove big businesses from businessList
 
+  }
+
+  public List<Listing> businesses() {
+    return businesses;
   }
 }
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -1,4 +1,5 @@
 package com.google.sps.data;
+import java.util.List;
 
 /** SmallCityService object representing all components of the webapp **/
 public class SmallCityService {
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/sps/data/SmallCityService.java
```diff
@@ -18,11 +20,21 @@ public SmallCityService(MapLocation mapLocation) {
...
   }
 
   public void eliminateBigBusinesses() {
     // TODO: Parse big business list and remove big businesses from businessList
 
   }
+
+  public List<Listing> businesses() {
+    return businesses;
+  }
```
ethanimooney, line 39: Should this be called something like getBusinesses() ?

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -14,29 +14,60 @@
...
+    return json;
   }
 
+  private double convertToDouble(String doubleAsString) {
+    double doubleAsDouble;
+    try {
+      doubleAsDouble = Double.parseDouble(doubleAsString);
+    } catch(NumberFormatException e) {
+      logger.warning("Location services failure - default set");
+      return 0; // Null Island
```
ethanimooney, line 69: nit: Should these be indented?

src/main/webapp/index.html
```diff
@@ -6,7 +6,8 @@
     <link href="https://fonts.googleapis.com/css2?family=Raleway:ital,wght@0,100;0,200;0,300;0,400;0,500;0,600;0,700;0,800;0,900;1,100;1,200;1,300;1,400;1,500;1,600;1,700;1,800;1,900&display=swap" rel="stylesheet">
     <title>smallcity</title>
   </head>
-  <body>
+  <!-- TODO: Move getGeolocation() function call to sensical location -->
```
ethanimooney, line 9: Don't let me forget to do this once we merge my branch into master, this may also create a small merge conflict for you.

src/main/java/com/google/sps/data/SmallCityService.java
LindseyWeiskopf: yes - just changed some of the naming conventions around to make more sense

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: They are on my screen - I changed them around a bit though. Does it look normal now?

src/main/webapp/script.js
```diff
@@ -12,3 +12,22 @@
...
+
+function displayLocation(position) {
+  var lat = position.coords.latitude;
+  var lng = position.coords.longitude;
+  var displayText = lat + ", " + lng;
+  document.getElementById("location").innerHTML = displayText;
+}
+
+function displayError() {
+  document.getElementById("location").innerHTML = "please enter your zipcode";
```
AbakirH: "please enter your zipcode" is repetitive in these lines of code, consider making a literal variable for this part of the string

src/main/webapp/script.js
```diff
@@ -30,5 +31,6 @@ function displayLocation(position) {
 }
 
 function displayError() {
-  document.getElementById("location").innerHTML = "please enter your zipcode";
+  // TODO: Write error to user interface
+  console.log("Error - please enter zipcode");
```
AbakirH, line 35: "please enter zipcode" is again repetitive here, possibly try to use literal variable for this. Since you deleted the previous code.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -15,35 +15,59 @@
...
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import java.util.logging.Logger;
 
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
   
   SmallCityService smallCityService;
+  private Logger logger = Logger.getLogger(DataServlet.class.getName());
```
AbakirH, line 36: logger is not really a descriptive variable name, what does it log?

src/main/webapp/script.js
```diff
@@ -12,3 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+function getGeolocation() {
+  if (navigator.geolocation) {
+    navigator.geolocation.getCurrentPosition(displayLocation, displayError);
+  }
+  else {
+    document.getElementById("location").innerHTML = "Your browser doesn't support geolocation - please enter zipcode";
```
shaquille-hall: Perhaps raise this as an alert in the browser (a try/catch should do, I believe)? Filling an html element with a warning like this may not be the best user experience.

src/main/webapp/script.js
LindseyWeiskopf: I updated the error message to be more specific - it's not user facing so I just described the issue instead of an action

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: I fixed the initalization to be a final static but I think LOGGER is descriptive enough because it can log anything and that is the conventional naming. Let me know if you disagree

AbakirH: I think the naming convention is weird for me since the variable name is the same as variable type. Like you would never do int int or double double, so maybe you can change it to printer

src/main/webapp/script.js
LindseyWeiskopf: Ah that's a great idea. I logged the issue to the console and created an alert that pops up on the top of the page. Let me know if this is better!

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: I know this is not the official style guide, but this is how the google development example handles this: https://cloud.google.com/logging/docs/setup/java#example_2

src/main/webapp/script.js
AbakirH: I was told it is good to never have large strings, assigned throughout the code, instead assign that string to a variable in the beginning of the code. Does this make sense? For example, when you are making a URL API call you should make a variable that you assign that string URL, rather then just writing that string throughout your code.

src/main/java/com/google/sps/servlets/DataServlet.java
AbakirH: sounds good then, thank you for the clarification

src/main/webapp/script.js
LindseyWeiskopf: I factored out the alert messages to a string literal. For now though, I  think I am going to keep the console messages inline because there are only used once and will likely become more specific. I don't think that they disrupt the flow of code too much.

src/main/webapp/index.html
shaquille-hall: @ethanimooney  Perhaps create an issue to track this?

src/main/java/com/google/sps/data/SmallCityService.java
```diff
@@ -12,17 +14,27 @@
...
   public SmallCityService(MapLocation mapLocation) {
     this.user = new User(mapLocation);
-    getAllBusinesses();
+    findAllBusinesses();
     eliminateBigBusinesses();
   }
 
-  public void getAllBusinesses() {
+  public void findAllBusinesses() {
     // TODO: Get businesses from Place API given user location
```
shaquille-hall, line 22: Please create issues for these todos.

PR #17, opened 2020-07-08
Implement ZipCodeToMapLocation function

LindseyWeiskopf:
Summary: Implement ZipCodeToMapLocation function with "Locate Me" button on the [front end](https://step2020-smallcity.appspot.com/)

Front end: I created a geolocate and zipcode form for users to prompt a location action. (This can 110% be changed by @ethanimooney)

Whether the user opts to be geolocated or enter a zipcode, it will call a javascript function. If the user enters a zipCode, the javascript will create a post request to send the zipcode to the servlet to create a user for the SmallCityService instance. If the user chooses to be geolocated, the getGeolocation() function will be called which uses the Geocoding API to convert the zipcode to a coordinate somewhere in the area designated by that zip. If the user allows geolocation from the browser, the function will create a post request for the servlet with the lat and lng coordinates to add to the smallCityService object.

I also remade the constructor of the SmallCityService class to avoid a null pointer exception in the Get Request.

NOTE: Right now, it appears that geolocation only works with https and not http. I'm not sure if this is an issue or if there is a solution to this.

### DIFF
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -8,16 +8,31 @@ public class SmallCityService {
   private User user;
   private List<Listing> businesses;
   
-  /** Create a new Small City Service instance
-  * @param mapLocation geolocation of user
-  * @return List of small local businesses
+  /** Create a new Small City Service instance from with no information **/
+  public SmallCityService() {
+
+  }
+  
+  /** 
+  * Create User instance from zipCode and get businesses list
+  * @param zipCode inputted zipcode of user
   **/
-  public SmallCityService(MapLocation mapLocation) {
+  public void createWithZip(String zipCode) {
+    this.user = new User(zipCode);
+    findAllBusinesses();
+    eliminateBigBusinesses();
+  }
+  
+  /** 
+  * Create User instance from geolocation and get businesses list
+  * @param mapLocation found geolocation of user
+  **/
+  public void createWithGeolocation(MapLocation mapLocation) {
     this.user = new User(mapLocation);
     findAllBusinesses();
     eliminateBigBusinesses();
   }
-
+  
   public void findAllBusinesses() {
     // TODO: Get businesses from Place API given user location
     businesses = new LinkedList<Listing>();
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -8,10 +8,7 @@ public class SmallCityService {
   private User user;
   private List<Listing> businesses;
   
-  /** Create a new Small City Service instance from with no information **/
-  public SmallCityService() {
-
-  }
+  public SmallCityService() { }
   
   /** 
   * Create User instance from zipCode and get businesses list
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -17,7 +17,7 @@ public class SmallCityService {
   * Create User instance from zipCode and get businesses list
   * @param zipCode inputted zipcode of user
   **/
-  public void createWithZip(String zipCode) {
+  public void createUserWithZip(String zipCode) {
     this.user = new User(zipCode);
     findAllBusinesses();
     eliminateBigBusinesses();
@@ -27,7 +27,7 @@ public class SmallCityService {
   * Create User instance from geolocation and get businesses list
   * @param mapLocation found geolocation of user
   **/
-  public void createWithGeolocation(MapLocation mapLocation) {
+  public void createUserWithGeolocation(MapLocation mapLocation) {
     this.user = new User(mapLocation);
     findAllBusinesses();
     eliminateBigBusinesses();
--- a/src/main/java/com/google/sps/data/User.java
+++ b/src/main/java/com/google/sps/data/User.java
@@ -15,7 +15,7 @@ import java.util.logging.Logger;
 public class User {
 
   private MapLocation geolocation;
-  private final String KEY = "AIzaSyDDIsG-SJAZ69ZoOecmfbXOB7ZIS4pZkAw";
+  private final String KEY = "REDACTED";
   private final static Logger LOGGER = Logger.getLogger(User.class.getName());
 
   /** Creates a user with a geolocation
--- a/src/main/java/com/google/sps/data/User.java
+++ b/src/main/java/com/google/sps/data/User.java
@@ -1,9 +1,22 @@
 package com.google.sps.data;
 
+import com.google.maps.GeocodingApi;
+import com.google.maps.model.AddressComponentType;
+import com.google.maps.model.AddressType;
+import com.google.maps.model.ComponentFilter;
+import com.google.maps.model.GeocodingResult;
+import com.google.maps.GeoApiContext;
+import com.google.maps.model.LatLng;
+import com.google.maps.model.LocationType;
+import java.util.Arrays;
+import java.util.logging.Logger;
+
 /** User representing a location **/
 public class User {
 
   private MapLocation geolocation;
+  private final String KEY = "AIzaSyDDIsG-SJAZ69ZoOecmfbXOB7ZIS4pZkAw";
+  private final static Logger LOGGER = Logger.getLogger(User.class.getName());
 
   /** Creates a user with a geolocation
   * @param geolocation lat/lng coordinate
@@ -17,11 +30,26 @@ public class User {
   * @param zipcode
   * @return User with zipCode
   **/
-  public User(int zipCode) {
-
+  public User(String zipCode) {
+    this.geolocation = zipCodeToMapLocation(zipCode);
   }
 
   public MapLocation getGeolocation() {
     return geolocation;
+  }
+
+  private MapLocation zipCodeToMapLocation(String zipCode) {
+    final GeoApiContext context = new GeoApiContext.Builder()
+            .apiKey(KEY)
+            .build();
+    final GeocodingResult[] results;
+    try {
+        results = GeocodingApi.geocode(context, zipCode).await();
+        MapLocation geolocation = new MapLocation(results[0].geometry.location.lat, results[0].geometry.location.lng);
+        return geolocation;
+    } catch (final Exception e) {
+        LOGGER.warning(e.getMessage());
+    }
+    return new MapLocation(0,0);  // null Island
   }
 }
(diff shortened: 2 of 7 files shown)


src/main/java/com/google/sps/data/SmallCityService.java
```diff
@@ -8,16 +8,31 @@
...
+  /** Create a new Small City Service instance from with no information **/
+  public SmallCityService() {
+
+  }
+  
+  /** 
+  * Create User instance from zipCode and get businesses list
+  * @param zipCode inputted zipcode of user
+  **/
+  public void createWithZip(String zipCode) {
```
ethanimooney, line 20: This is definitely a nit, and only if you want - you might change createWithZip() and createWithGeolocaton() to createUserWith... . Might make it a bit more clear for anyone looking at this for the first time

src/main/java/com/google/sps/data/User.java
```diff
@@ -17,11 +30,26 @@ public User(MapLocation geolocation) {
...
   }
 
   public MapLocation getGeolocation() {
     return geolocation;
   }
+
+  private MapLocation zipCodeToMapLocation(String zipCode) {
+    final GeoApiContext context = new GeoApiContext.Builder()
+            .apiKey(KEY)
+            .build();
```
ethanimooney, line 44: Is there a reason these lines are indented like this?

src/main/webapp/index.html
```diff
@@ -7,11 +7,16 @@
     <title>smallcity</title>
   </head>
   <!-- TODO: Move getGeolocation() function call to sensical location -->
```
ethanimooney, line 9: Can probably remove this now?

ethanimooney: The comment, that is

src/main/java/com/google/sps/data/SmallCityService.java
LindseyWeiskopf: Ah yes that's a lot clearer. Just changed.

src/main/java/com/google/sps/data/User.java
LindseyWeiskopf: The [style guide](https://google.github.io/styleguide/javaguide.html#s4.1-braces) says to use 4+ spaces when separating by the dot operator. I think it just makes it easier to read

src/main/webapp/index.html
LindseyWeiskopf: It's gone!

src/main/java/com/google/sps/data/User.java
ethanimooney: Ah okay, cool

src/main/java/com/google/sps/data/User.java
```diff
@@ -17,11 +30,26 @@ public User(MapLocation geolocation) {
...
+    final GeocodingResult[] results;
+    try {
+        results = GeocodingApi.geocode(context, zipCode).await();
+        MapLocation geolocation = new MapLocation(results[0].geometry.location.lat, results[0].geometry.location.lng);
```
shaquille-hall, line 48: Should this only care about the first element in the results?

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -32,19 +32,22 @@
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
```
shaquille-hall, line 34: Please add a comment describing the responsibility of this class.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -32,19 +32,22 @@
...
     String lngString = request.getParameter("lng");
     double lat = convertToDouble(latString);
     double lng = convertToDouble(lngString);
-    // Google Pittsburgh Office Location (hardcoded prototype)
-    // Will be populated with lat and lng doubles
-    MapLocation userLocation = new MapLocation(40.457410, -79.916573);
-    smallCityService = new SmallCityService(userLocation);
+    MapLocation userLocation = new MapLocation(lat, lng);
+    smallCityService.createUserWithGeolocation(userLocation);
+    } catch(NullPointerException e) {
```
shaquille-hall, line 47: Where is the NPE getting thrown in this block? Is it request.getParameter()?

src/main/webapp/script.js
```diff
@@ -33,7 +34,13 @@ function displayLocation(position) {
 }
 
 function displayError() {
-  // TODO: Write error to user interface
   console.log('Geolocation not enabled');
   alert(alertMessage);
+}
+
+function getZip() {
```
shaquille-hall, line 41: *getZipCode(). There's no real need to short hand it.

src/main/webapp/script.js
```diff
@@ -33,7 +34,13 @@ function displayLocation(position) {
...
+  var zip = document.getElementById('zipCode').value;
+  var xhttp = new XMLHttpRequest();
```
shaquille-hall, line 43: Reminder, we decided to prefer const or let.

src/main/java/com/google/sps/data/SmallCityService.java
```diff
@@ -8,16 +8,31 @@
   private User user;
   private List<Listing> businesses;
   
-  /** Create a new Small City Service instance
-  * @param mapLocation geolocation of user
-  * @return List of small local businesses
+  /** Create a new Small City Service instance from with no information **/
+  public SmallCityService() {
+
+  }
```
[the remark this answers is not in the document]
LindseyWeiskopf: Okay, that makes a lot of sense - I removed the comment and fixed the brackets.

src/main/java/com/google/sps/data/User.java
LindseyWeiskopf: The API documentation says that generally only one result is returned unless the input is particularly ambiguous. It seems as though every zipCode that I have tried only returns one result.

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: Done!

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: Yes

src/main/webapp/script.js
LindseyWeiskopf: Got it - it's changed

src/main/webapp/script.js
LindseyWeiskopf: Ah, yes gotta love Javascript!

src/main/java/com/google/sps/data/User.java
shaquille-hall: I couldn't the docs for this method? Does it provide a getFirst() call or similar?

src/main/java/com/google/sps/servlets/DataServlet.java
shaquille-hall: Gotcha. What happens if geolocation isn't enabled in the browser, and the zipcode isn't entered? Is this possible? Do you prevent proceeding for users who haven't entered any of these? 

Also, it's a good idea to log something in your catch block here.

src/main/java/com/google/sps/data/User.java
LindseyWeiskopf: Not that I know of. This is the class that the method is from: https://github.com/googlemaps/google-maps-services-java/blob/master/src/main/java/com/google/maps/GeocodingApi.java

I was just referencing the location by the first element in the array. Does that answer your question?

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: Right now, if the user presses submit without entering a zipcode, nothing happens (no results). I think this will be more clearer from a UI standpoint once the popup is added when the user enters the site.

src/main/java/com/google/sps/data/User.java
shaquille-hall: Yes. But looking at this method, it doesn't seem to return multiple objects. Just one.

src/main/java/com/google/sps/servlets/DataServlet.java
shaquille-hall: Ok. Adding a safeguard in the frontend isn't a bad idea.

Please add a log below outlining why the NPE here is ok.

src/main/java/com/google/sps/data/User.java
LindseyWeiskopf: I think that the .await() triggers the getResult() function (along with the others) at the end of that file which does return an array of results. I tried to make the type just a GeocodingResult in my code and I got an incompatible types error.

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: I added a logger message and created an issue to track the frontend safeguard #24

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -29,22 +29,28 @@
 import javax.servlet.http.HttpServletResponse;
 import java.util.logging.Logger;
 
+/** Servlet that recieves user location and sends small business output to the client **/
```
shaquille-hall, line 32: nit: You can omit "servlet that"

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -29,22 +29,28 @@
...
-    // Will be populated with lat and lng doubles
-    MapLocation userLocation = new MapLocation(40.457410, -79.916573);
-    smallCityService = new SmallCityService(userLocation);
+    MapLocation userLocation = new MapLocation(lat, lng);
+    smallCityService.createUserWithGeolocation(userLocation);
+    } catch(NullPointerException e) {
+      String zip = request.getParameter("zipCode");
+      smallCityService.createUserWithZip(zip);
+      LOGGER.warning(e.getMessage());
+      LOGGER.warning("Unable to geolocate user, zipCode entered instead.");
```
shaquille-hall, line 52: nit: Can this be converted to use string concatenation for just one log statement?

src/main/java/com/google/sps/data/User.java
shaquille-hall: Gotcha. I see now. Thanks for clarifying this.

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: Done!

PR #21, opened 2020-07-09
Add frontend list view with pagination cababilities

ethanimooney:
TLDR; Added the main list view to the frontend, has pagination abilities as well.
LINK: https://step2020-smallcity.appspot.com/

**NOTE: after committing I am realizing I need to rename displayList() and limitList() to more accurately represent their functions. Will do in next commit after initial reviews**

**index.html** : Removed demo cards, added displayList() call to body onLoad, created a new container inside of results-continer called results-content. This was done to separate the container to append the cards to from the button container to avoid issues associated. The page buttons were also changed to inputs with a type of button.

**script.js** : 
- Changed all _var_ to _let_. 

- **displayList()** This function fetches the data from _DataServlet.java_ and iterates through the returned JSON, calling the createResultCard() function, creating the individual cards for results and then appending them to the listingsArray, then calls the limitList() function which takes the cards in listingsArray and prints them to the html.

- **createResultCard()** This function is passed the individual name, address, etc. of a listing from the JSON and creates the individual divs and then appends them to their particular resultsCard. It then returns that resultsCard back to displayList().

- **createRating()** This function is called in createResultCard() when adding the rating information. It takes in the rating double, rounds it to the nearest integer, and then displays filled in / not filled in stars accordingly. It appends these stars to the ratingDiv and also adds the numerical value for users to see. This ratingDiv is sent back to createResultCard() and is appended to the card.

- **limitList()** This function is what controls the pagination and actual displaying of the list view. It takes in a variable called direction, which is either _begin, previous, next_ . This determines the action that the function will take. If passed _begin_, the function displays the first 3 cards in listingsArray, setting the currentEndCard to 2 (the third card). If passed _previous_, the function uses the currentEndCard value to move back 5 spaces, and then appends the next 3 cards. This prints the three cards before the current first card. If passed _next_, the function uses the currentEndCard value to print out the next 3 cards after the current end card. All of this is preceded by an if statement checking whether or not the user is already at the end or beginning of the list. If so, it does not let them 'go off the end or beginning' and (currently) displays an alert.

**style.css** :  Changed CSS to reflect the changes in index.html. I.E. changing the CSS-Grid layout for results-container, adding styling for results-content, few other minor style changes.


src/main/webapp/index.html
```diff
@@ -7,7 +7,7 @@
     <title>smallcity</title>
   </head>
   <!-- TODO: Move getGeolocation() function call to sensical location -->
-  <body onload="getGeolocation()">
+  <body onload="getGeolocation(); displayList()">
```
LindseyWeiskopf, line 10: The Location of the display function (and getGeolocation function) will likely have to move to make sure that we do not try to display any listings before we have the user's location. I'll create an issue to track this

src/main/webapp/script.js
```diff
@@ -36,4 +36,124 @@ function displayError() {
   // TODO: Write error to user interface
   console.log('Geolocation not enabled');
   alert(alertMessage);
+}
+
+//Array of the (currently 6 for this demo build) 15 listings gathered from the fetch request
+const listingsArray = [];
```
LindseyWeiskopf, line 42: Should this be a 'let' since elements will be added to the array?

src/main/webapp/script.js
```diff
@@ -36,4 +36,124 @@ function displayError() {
...
+}
+
+//Variable keeping track of the card that is currently shown at the bottom of the list
+let currentEndCard;
+
+/**
+ * @param {string} direction Contains which direction the list should iterate
+ */
+function limitList(direction) {
+  const resultsContent = document.getElementById('results-content');
```
LindseyWeiskopf, line 62: Same here as above

src/main/webapp/script.js
```diff
@@ -36,4 +36,124 @@ function displayError() {
...
+  resultsContent.className = 'results-content';
+
+  // TODO: Add better checking for end / beginning of list and change/remove alert
+  // TODO: Also add changing nav buttons based on this checking
+  if((currentEndCard - 2) == 0 && direction == 'previous') {
```
LindseyWeiskopf, line 67: A quick explanation comment may be helpful here.

src/main/webapp/script.js
```diff
@@ -36,4 +36,124 @@ function displayError() {
...
+    alert('Already at beginning of list!');
+  }
+  else if ((currentEndCard + 1) == count && direction == 'next') {
+    alert('Already at end of list!');
+  }
+  else {
+    resultsContent.innerHTML = '';
+    if(direction == 'begin') {
+      for (i = 0; i < 3; i++) {
```
LindseyWeiskopf, line 76: Do you declare 'i' anywhere? Not sure if that will cause issues with the javascript down the road

src/main/webapp/script.js
ethanimooney: Yes, good catch

src/main/webapp/script.js
ethanimooney: I'm not sure here. This is how the starter project example code did it, so I kept with this method

shaquille-hall: Are we referring to const vs. let? If so, const seems fine here.

src/main/webapp/script.js
```diff
@@ -36,4 +36,124 @@ function displayError() {
...
+  });
+}
+
+//Variable keeping track of the card that is currently shown at the bottom of the list
+let currentEndCard;
+
+/**
+ * @param {string} direction Contains which direction the list should iterate
+ */
+function limitList(direction) {
```
shaquille-hall, line 61: If you're interested, you can move the pagination bits to a different script. It might allow for improved modularity, and parsing later on.

src/main/webapp/index.html
```diff
@@ -16,30 +16,12 @@ <h1 class="logo">smallcity</h1>
...
-          
-        </div>
+      <div class="map-container"></div>
+      <div class="results-container" id="results-container">
+       <div class="results-content" id="results-content"></div>
         <div class="pagination-container">
-          <button class="pagination-buttons">Previous</button>
-          <button class="pagination-buttons">Next</button>
+          <input type="button" class="pagination-buttons" value="Previous" onclick="limitList('previous')">
+          <input type="button" class="pagination-buttons" value="Next" onclick="limitList('next')">
```
shaquille-hall, line 24: I'd suggest refactoring this into two methods: navigatePrevious and moveNext or something similar. That makes things a bit clearer, and will remove the need for some of if statements below.

src/main/webapp/script.js
ethanimooney: Oops haha - dunno how that was working

src/main/webapp/index.html
ethanimooney: Gotcha

src/main/webapp/index.html
ethanimooney: That's a great idea. I did it this way originally to avoid confusion but ended up making it more complex haha - changed

src/main/webapp/script.js
ethanimooney: Good idea, doing that now.

ethanimooney:
This commit addresses all the above comments, adds the paginationScript.js file for all functions related, and also adds /images and smallcity-favicon-15x.png.

src/main/webapp/paginationScript.js
```diff
@@ -0,0 +1,63 @@
...
+//Display the initial 3 cards in the list
+function initialDisplay() {
+  const resultsContent = document.getElementById('results-content');
+  resultsContent.className = 'results-content';
+
+  for (let i = 0; i < 3; i++) {
+    resultsContent.appendChild(listingsArray[i]);
+  }
+  currentEndCard = 2;
+}
```
shaquille-hall, line 27: You can generalize this method by passing in a parameter to determine how many cards to display at a time. With a few more lines, this can become pretty reusable. Think:

moveNext() {
....
displayCards
...
}

navigatePrevious() {
....
displayCards
...
}

ethanimooney:
Added displayCards() function

ethanimooney:
This commit fixes merge conflicts, and removes the fetchList() call from the body onload, so currently the list will not display, but I am working with lweiskopf@ to figure out where to put it.

ethanimooney:
I did not push this commit to the deployed site

src/main/webapp/script.js
```diff
@@ -38,6 +38,87 @@ function displayError() {
...
+//Count of the total businesses in the fetch request, used to set a unique id for each card
+let count = 0;
+
+function fetchList() {
+  fetch('/data').then(response => response.json()).then((listings) => {
+    listings.forEach((listing) => {
+      listingsArray.push(createResultCard(listing.name, listing.address, listing.image, listing.rating, listing.website, count));
+      count++;
+    });
+    initialDisplay();
```
shaquille-hall, line 53: Does this need to be in the fetch? Can the list be fetched, and the cards be created, then displayed?

ethanimooney:
I addressed the change you recommended, shaquilleh@, and also changed **count** to **listingCardCount** for more clarity on what the variable does. This commit has also not been pushed to the live app, for reasons that you know of already

src/main/webapp/paginationScript.js
```diff
@@ -0,0 +1,51 @@
...
+  else {
+    displayCards((resultsContent.children[2].id + 1))
+  }
+}
+
+//Handles the actual loop through array and display of cards
+function displayCards(starting) {
+  resultsContent.innerHTML = '';
+
+  for (let i = starting; i < (starting + 3); i++) {
```
shaquille-hall, line 48: Please make '3' a const .

src/main/webapp/script.js
```diff
@@ -38,6 +38,87 @@ function displayError() {
   alert(alertMessage);
 }
 
+//Array of the (currently 6 for this demo build) 15 listings gathered from the fetch request
+let listingsArray = [];
+
+//Count of the total businesses in the fetch request, used to set a unique id for each card
+let listingCardCount = 0;
```
shaquille-hall, line 45: nit: Just a suggestion: I'd rename this "totalCardCount" specifically based on how it's being used.

src/main/webapp/paginationScript.js
```diff
@@ -0,0 +1,51 @@
...
+resultsContent.className = 'results-content';
+
+//Display the initial 3 cards in the list
+function initialDisplay() {
+  displayCards(0);
+}
+
+//Navigate back one page (3 results) in the list view
+function navigatePrevious() {
+  if (resultsContent.children[0].id == 0) {
```
shaquille-hall, line 26: I admit, this is a pretty creative solution. Are you familiar with the indexOf() method? This would produce a more contained solution in js.

src/main/webapp/paginationScript.js
```diff
@@ -0,0 +1,51 @@
...
+    alert('Already at beginning of list!');
+  }
+  else {
+    displayCards((resultsContent.children[2].id - 5))
```
shaquille-hall, line 30: Why 5?

ethanimooney: Ohhh wait I'm making that harder than it should be. I should do resultsContent.children[0].id - 3

I was going from the current last position to the next first position.

src/main/webapp/script.js
ethanimooney: Done

src/main/webapp/paginationScript.js
ethanimooney: Got it

ethanimooney:
TLDR; Got front end working with lweiskopf@, changed pagination implementation as suggested by shaquilleh.
LINK: https://step2020-smallcity.appspot.com

**To view the display in action, press locate me or enter a zip code.**

I have currently limited the amount of results displayed to 15.

 There are a few backend changes, for answers on those I would ask lweiskopf@, she will be able to describe better.

Changes in index or script were also just to get the frontend displaying.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -52,9 +52,12 @@ public BusinessesService(List<Listing> businesses) {
         for(PlacesSearchResult place : response.results) {
           addListingToBusinesses(place);
         }
-        Thread.sleep(2000); // Required delay before next API request
-	      response = PlacesApi
-              .nearbySearchNextPage(context, response.nextPageToken).await();
+        //Maximum of 2 next token requests allowed
```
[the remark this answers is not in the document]
LindseyWeiskopf: Just popping in to respond to this one -- the API[ documentation](https://developers.google.com/places/web-service/search#PlaceSearchPaging) says that you can only make a maximum of 2 requests to the nextPageToken function (for a total of 60 results) that are called a short delay apart from each other. If the function is called more than 2 times, it returns a null point exception that kept the listings from displaying. Granted, this if statement is not the most elegant approach to this issue, but it was a quick fix to keep the webapp from breaking.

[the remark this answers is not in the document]
LindseyWeiskopf: #27

src/main/webapp/paginationScript.js
```diff
@@ -0,0 +1,60 @@
...
+
+//Move forward one page (3 results) in the list view
+function moveNext() {
+  displayCards(3);
+}
+
+//Handles the actual loop through array and display of cards
+function displayCards(listAugment) {
+  const totalCardsToDisplay = 3;
+  const maxListViewNumber = 15;
```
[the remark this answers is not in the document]
ethanimooney: Great, done

src/main/webapp/script.js
ethanimooney: It was, but we found that the function would actually call the initialDisplay() function before doing the fetch, causing the whole thing to break, so I moved it back inside the fetch.

src/main/java/com/google/sps/data/BusinessesService.java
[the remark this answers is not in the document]
shaquille-hall: Thanks for clarifying that. Please add this description to the issue. 

I think the if statement is fine. Although, this is somewhat fragile. What happens if ALLOWED_SEARCH_REQUESTS becomes anything greater than 3?

src/main/webapp/paginationScript.js
```diff
@@ -0,0 +1,60 @@
...
+
+  let resultsChildren = resultsContent.childNodes;
```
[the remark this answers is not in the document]
ethanimooney: Yes, I moved it out of displayCards() just into the main body of paginationScript.js

src/main/webapp/paginationScript.js
```diff
@@ -0,0 +1,60 @@
...
+  else if (currentFirstCardIndex == (maxListViewNumber - totalCardsToDisplay) && (listAugment > 0)) {
+    alert('Already at end of list!');
+  }
+  else {
+    currentFirstCardIndex += listAugment; 
+
+    resultsContent.innerHTML = '';
+    for (let i = currentFirstCardIndex; i < (currentFirstCardIndex + totalCardsToDisplay); i++) {
+      resultsContent.appendChild(listingsArray[i]);
+    }
```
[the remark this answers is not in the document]
shaquille-hall: Apologies. I misread this block. You can disregard this comment.

src/main/webapp/script.js
```diff
@@ -31,16 +31,102 @@ function displayLocation(position) {
...
+
+  const nameHeader = document.createElement('h2');
+  nameHeader.innerText = name;
+  const addressParagraph = document.createElement('p');
+  addressParagraph.innerText = address;
+
+  const nameAndAddressDiv = document.createElement('div');
+  nameAndAddressDiv.className = 'results-business-description';
+  nameAndAddressDiv.appendChild(nameHeader);
+  // TODO: Actually add address once lweiskopf@ pr is approved, is just 'undefined' for now
```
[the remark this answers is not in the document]
ethanimooney: This comment should have been removed, it's gone now.

ethanimooney:
Addressed the above new comments.

src/main/java/com/google/sps/data/BusinessesService.java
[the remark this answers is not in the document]
LindseyWeiskopf: #27

PR #25, opened 2020-07-10
User class Test

LindseyWeiskopf:
Summary: This is a test class for the user object, primarily to test the zipCodeToGeolocation Location function.

There are three main tests in this file:

**geolocationUserWithMapLocation():** This test checks to make sure that the geolocation returned by the getGeolocation() function is the same as the entered map location when the user is located by the browser.

**geolocationUserWithZipCode():** This test checks the functionality of the zipCodeTOGeolocation function when the user enters a zipcode. The test is determined by if the coordinate returned by the Geolocation API falls within the general range of the example zipcode. I found these values by finding the differences in lat and lng of opposite corners of the zipCode range.

**geolocationUserWithCity():** This test works similarly to the above but when the user enters a city. I also found the coordinate range of the opposite coordinates to create a range of acceptable responses.

### DIFF
--- a/src/test/java/com/google/sps/UserTest.java
+++ b/src/test/java/com/google/sps/UserTest.java
@@ -21,7 +21,6 @@ public final class UserTest {
   private final MapLocation USER_MAP_LOCATION = new MapLocation(40.45717, -79.91669);
   private final String USER_ZIPCODE = "15206";
   private final String USER_CITY = "Pittsburgh";
-  private final MapLocation NULL_ISLAND = new MapLocation(0, 0);
   
   private User userWithMapLocation;
   private User userWithZipCode;
--- a/src/test/java/com/google/sps/UserTest.java
+++ b/src/test/java/com/google/sps/UserTest.java
@@ -27,10 +27,12 @@ import java.util.logging.Logger;
 @RunWith(JUnit4.class)
 public final class UserTest {
   // Pittsburgh Google Office Location
-  private final MapLocation TEST_USER_MAP_LOCATION = new MapLocation(40.45717, -79.91669);
+  private final MapLocation TEST_USER_MAP_LOCATION 
+        = new MapLocation(40.45717, -79.91669);
   private final String TEST_USER_ZIPCODE = "15206";
   private final String TEST_USER_CITY = "Pittsburgh";
-  private final String TEST_USER_ADDRESS = "6425 Penn Ave, Pittsburgh, PA 15206";
+  private final String TEST_USER_ADDRESS 
+        = "6425 Penn Ave, Pittsburgh, PA 15206";
   
   private User test_userWithMapLocation;
   private User test_userWithZipCode;
@@ -38,7 +40,8 @@ public final class UserTest {
   private User test_userWithAddress;
 
   private final String KEY = "REDACTED";
-  private final static Logger LOGGER = Logger.getLogger(UserTest.class.getName());
+  private final static Logger LOGGER 
+        = Logger.getLogger(UserTest.class.getName());
 
   @Test
   public void geolocationUserWithMapLocation() {
@@ -71,8 +74,9 @@ public final class UserTest {
     String address = reverseGeocode(actual); 
     Assert.assertTrue(address.contains(TEST_USER_ADDRESS));
   }
-
-  public String reverseGeocode(MapLocation mapLocation) {
+  
+  /** Returns formatted address from coordinate using Geocoding API **/
+  private String reverseGeocode(MapLocation mapLocation) {
     final GeoApiContext context = new GeoApiContext.Builder()
             .apiKey(KEY)
             .build();
@@ -85,6 +89,6 @@ public final class UserTest {
     } catch (final Exception e) {
         LOGGER.warning(e.getMessage());
     }
-    return "";
+    return "FAILURE";
   }
 }
--- a/src/test/java/com/google/sps/UserTest.java
+++ b/src/test/java/com/google/sps/UserTest.java
@@ -3,16 +3,21 @@ package com.google.sps;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.lang.Math;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
+import com.google.sps.data.SmallCityService;
+import com.google.sps.data.Listing;
+import com.google.sps.data.MapLocation;
+import com.google.sps.data.User;
 
 @RunWith(JUnit4.class)
 public final class UserTest {
   // Pittsburgh Google Office Location
-  private final MapLocation USER_MAP_LOCATION = new MapLocation(40.457177, -79.916696);
+  private final MapLocation USER_MAP_LOCATION = new MapLocation(40.45717, -79.91669);
   private final String USER_ZIPCODE = "15206";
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,84 @@
...
+  // Pittsburgh Google Office Location
+  private final MapLocation USER_MAP_LOCATION = new MapLocation(40.45717, -79.91669);
+  private final String USER_ZIPCODE = "15206";
+  private final String USER_CITY = "Pittsburgh";
+  private final String USER_ADDRESS = "6425 Penn Ave, Pittsburgh, PA 15206";
+  
+  private User userWithMapLocation;
+  private User userWithZipCode;
+  private User userWithCity;
+  private User userWithAddress;
```
shaquille-hall, line 29: nit: Typically we prefix these with "TEST_" and "test_" respectively

src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+  @Before
+  public void setUp() {
```
shaquille-hall, line 32: setup() is often useful, but not always required. Here it doesn't add much value, and the tests below become more readable if these definitions are included there, where relevant. 
e.g. 

```
@Test
  public void geolocationUserWithMapLocation() {
    userWithMapLocation = new User(USER_MAP_LOCATION);
   .................
  }
```

src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,84 @@
...
+    Assert.assertEquals(expected, actual);
+  }
+
+  @Test
+  public void geolocationUserWithZipCode() {
+    MapLocation actual = userWithZipCode.getGeolocation();
+    MapLocation expected = USER_MAP_LOCATION;
+    boolean inRange = false;
+    // Approximate range of lat and lng coordinates in 15206 ZipCode
+    if ((Math.abs(actual.lat - expected.lat) <= 0.06)
```
shaquille-hall, line 52: What is the significance of 0.06 and 0.05?

src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  @Test
+  public void geolocationUserWithCity() {
+    MapLocation actual = userWithCity.getGeolocation();
+    MapLocation expected = USER_MAP_LOCATION;
+    boolean inRange = false;
+    // Approximate range of lat and lng coordinates in Pittsburgh
+    if ((Math.abs(actual.lat - expected.lat) <= 0.16) 
+          && (Math.abs(actual.lng - expected.lng)) <= 0.2) {
```
shaquille-hall, line 66: Same comment as above. What's the significance of these values?

src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+  @Test
+  public void geolocationUserWithAddress() {
+    MapLocation actual = userWithAddress.getGeolocation();
+    MapLocation expected = USER_MAP_LOCATION;
+    boolean inRange = false;
+    // Small range of error allowed for address geolocation
+    if ((Math.abs(actual.lat - expected.lat) <= 0.001) 
+          && (Math.abs(actual.lng - expected.lng)) <= 0.001) {
+      inRange = true;
```
shaquille-hall, line 80: Ditto.

src/test/java/com/google/sps/UserTest.java
LindseyWeiskopf: just changed.

src/test/java/com/google/sps/UserTest.java
LindseyWeiskopf: Okay this makes a lot of sense

src/test/java/com/google/sps/UserTest.java
LindseyWeiskopf: I took the coordinates of the corners of the zipcodes and these values are the approximate ranges, so theoretically any coordinate within this difference will be in this zipcode. Kevin and I discussed this as a solution for testing the API results yesterday during our 1:1, but I'd love to hear any other ideas that you have.

src/test/java/com/google/sps/UserTest.java
LindseyWeiskopf: Same idea but for the city of Pittsburgh

src/test/java/com/google/sps/UserTest.java
LindseyWeiskopf: These are just very small margins of error that could happen for a specific address. I updated the comments - let me know if they still need clarifying.

LindseyWeiskopf:
I changed the tests so that the geocoding results are validated through a reverse geocoding call to the Geocoding API

src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+  @Test
+  public void geolocationUserWithAddress() {
+    test_userWithAddress = new User(TEST_USER_ADDRESS);
+    MapLocation actual = test_userWithAddress.getGeolocation();
+    String address = reverseGeocode(actual); 
+    Assert.assertTrue(address.contains(TEST_USER_ADDRESS));
+  }
+
+  public String reverseGeocode(MapLocation mapLocation) {
```
shaquille-hall, line 75: Does this need to be public?
Please add a comment about the purpose of this method. It may not be entirely clear to a third party reading it.

src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,90 @@
...
+    final GeocodingResult[] results;
+    final LatLng latLng = new LatLng(mapLocation.lat, mapLocation.lng);
+    try {
+        results = GeocodingApi.reverseGeocode(context, latLng).await();
+        String address = results[0].formattedAddress;
+        return address;
+    } catch (final Exception e) {
+        LOGGER.warning(e.getMessage());
+    }
+    return "";
```
shaquille-hall, line 88: The test will fail when this returns empty one day.

src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,90 @@
...
+import com.google.maps.GeoApiContext;
+import com.google.maps.model.LatLng;
+import com.google.maps.model.LocationType;
+import java.util.logging.Logger;
+
+/** Test class for User Object **/
+@RunWith(JUnit4.class)
+public final class UserTest {
+  // Pittsburgh Google Office Location
+  private final MapLocation TEST_USER_MAP_LOCATION = new MapLocation(40.45717, -79.91669);
```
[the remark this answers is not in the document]
LindseyWeiskopf: No...but it is now!

src/test/java/com/google/sps/UserTest.java
LindseyWeiskopf: Changed to private and added comment about the function

src/test/java/com/google/sps/UserTest.java
LindseyWeiskopf: True. I changed the return value to "FAILURE" -- hopefully there is not an address called failure somewhere

shaquille-hall: It wasn't an issue. I was just making a note. But this is more descriptive, and indicates the known failure case. It works.

PR #26, opened 2020-07-10
Fix conflict between pull requests

LindseyWeiskopf:
When I merged my pull requests, the zipCode constructor didn't envoke the BigBusinesses class so the functionality did not work. I added that call to the zipCode constructed and refactored the shared code. Thanks!

### DIFF
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -33,7 +33,7 @@ public class SmallCityService {
   **/
   public void createUserWithZip(String zipCode) {
     this.user = new User(zipCode);
-    createBusinessesService();
+    getSmallBusinesses();
   }
   
   /** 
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/sps/data/SmallCityService.java
```diff
@@ -63,14 +58,10 @@ public void eliminateBigBusinesses() {
...
-  private void addListingToBusinesses(PlacesSearchResult place) {
-    String name = place.name;
-    String formattedAddress = place.vicinity;
-    Geometry geometry = place.geometry;
-    MapLocation placeLocation = new MapLocation(geometry.location.lat, geometry.location.lng);
-    double rating = place.rating;
-    Photo photos[] = place.photos;
-    String types[] = place.types;
-    businesses.add(new Listing(name, formattedAddress, placeLocation, rating, photos, types));
+  private void createBusinessesService() {
```
shaquille-hall, line 61: Rename this to focus on the goal from the application's pov, not the implementation. Something like "getSmallBusinesses()" etc..

LindseyWeiskopf: Changed! Thank you!

PR #29, opened 2020-07-14
Map and markers

LindseyWeiskopf:
Summary: Add map to frontend and create markers for each listing
[webapp](https://step2020-smallcity.appspot.com/)

After the list of businesses is returned, I create a map instance using the Maps API centered around the location of the first listing returned.

When parsing through all of the listings, I have a createMarker function that creates a marker for each listing on the map.

The styling change allows the map corners to be rounded. I changed the map from a div to an id to accommodate the API

Thanks!

### DIFF
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -57,7 +57,6 @@ function fetchList() {
       createMarker(listing);
     });
     initialDisplay();
-    
   });
 }
 
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -53,7 +53,6 @@ function fetchList() {
     totalCardCount = 0;
     initMap(listings[0].mapLocation);
     listings.forEach((listing) => {
-      console.log(totalCardCount);
       listingsArray.push(createResultCard(listing.name, listing.formattedAddress, listing.photos, listing.rating, totalCardCount));
       totalCardCount++;
       if (totalCardCount < 15) createMarker(listing);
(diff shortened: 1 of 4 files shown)


src/main/webapp/script.js
```diff
@@ -129,4 +132,20 @@ function getZipCode() {
   xhttp.send();
 
   setTimeout(fetchList, 4000);
+}
+
+function initMap(mapLocation) {
```
ethanimooney, line 137: Let's move the map functions into their own .js file

shaquille-hall: Agreed.

src/main/webapp/script.js
```diff
@@ -129,4 +132,20 @@ function getZipCode() {
...
+  let myLatLng = {lat: mapLocation.lat, lng: mapLocation.lng};
+  map = new google.maps.Map(document.getElementById('map'), {
+    zoom: 13,
+    center: myLatLng,
+    mapTypeControl: false
+  });
+}
+
+function createMarker(listing) {
```
shaquille-hall, line 146: There seems to be a discrepancy between the number of marker shown, and the number of results shown. Perhaps these should be the same?

src/main/webapp/script.js
LindseyWeiskopf: It has been moved!

src/main/webapp/script.js
LindseyWeiskopf: Now the map should only show the 15 markers that appear as listings.

LindseyWeiskopf: [deployed here!](https://step2020-smallcity.appspot.com)

PR #30, opened 2020-07-15
A way to use API keys safely in our application

AbakirH:
Summary: Using the dotenv method to hide any API keys or other important information we may want 

Added the name of the file to .gitignore so we do not accidentally push the file to GitHub.
Added the appropriate code to use the Places API key throughout the code.

### DIFF
--- a/src/main/java/com/google/sps/data/BusinessesService.java
+++ b/src/main/java/com/google/sps/data/BusinessesService.java
@@ -34,7 +34,7 @@ public class BusinessesService {
   private Dotenv dotenv = Dotenv.configure().load();
 
   private List<Listing> allBusinesses;
-  private final String KEY = dotenv.get("MapsAPI");
+  private final String KEY = dotenv.get("PlacesAPI");
   private final static Logger LOGGER = 
         Logger.getLogger(BusinessesService.class.getName());
   private final int ALLOWED_SEARCH_REQUESTS = 3;
--- a/src/main/java/com/google/sps/data/BusinessesService.java
+++ b/src/main/java/com/google/sps/data/BusinessesService.java
@@ -25,14 +25,16 @@ import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.PreparedQuery;
 import com.google.appengine.api.datastore.Query;
 import com.google.appengine.api.datastore.Query.SortDirection;
+import io.github.cdimascio.dotenv.Dotenv;
 
 /** BusinessesService object representing all businesses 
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -25,14 +25,16 @@
...
 /** BusinessesService object representing all businesses 
 * components of the webapp.
  **/
 public class BusinessesService {
-  
+  private Dotenv dotenv = Dotenv.configure().load();
+
   private List<Listing> allBusinesses;
-  private final String KEY = "REDACTED";
+  private final String KEY = dotenv.get("PlacesAPI");
```
LindseyWeiskopf, line 37: Super nit: Since the key can be used for any API throughout the webapp, a more general name might be helpful in the future

src/test/java/com/google/sps/UserTest.java
```diff
@@ -22,10 +22,13 @@
 import com.google.maps.model.LatLng;
 import com.google.maps.model.LocationType;
 import java.util.logging.Logger;
+import io.github.cdimascio.dotenv.Dotenv;
```
LindseyWeiskopf, line 25: I'm not sure how to go about doing this, but we might want to look into the licensing for this to make sure that we're not violating any open source policy.

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Isn't this key specifically only for the places or maps feature. Like I am assuming a different api key is for the Custom Search API for example

src/test/java/com/google/sps/UserTest.java
AbakirH: It has a MIT license and I believe that should be okay for us to use

src/main/java/com/google/sps/data/BusinessesService.java
LindseyWeiskopf: We can use the same API key across the whole application for any Google API. I think the best practice is not share keys across different apps, so we should be okay.

AbakirH: Yes you are right, the same Key is for all the different API's Google offers. I just changed the name of it, to APIKey if that is better

src/test/java/com/google/sps/UserTest.java
shaquille-hall: The license permitting commercial use is here

PR #36, opened 2020-07-16
Test the filterBySmallBusinesses method 

AbakirH:
Summary: Removed a line of code from testFilterBySmallBusinessesMethod function, that would connect to the places API and get results from there. The test should only be based on the sample list and only test the one BusinessesService class. Also, added the correct address to each location for the sample businesses list.


AbakirH:
> The description mentions a line was removed from "testFilterBySmallBusinessesMethod function" but it doesn't look like it's here. Was it missed in the last commit?

I you press the first commit message, it should show you the line that was deleted, commit bffd337.

src/test/java/com/google/sps/BigBusinessesTest.java
```diff
@@ -155,16 +155,15 @@ public void testSampleDatabase() {
   }
 
   @Test
-  public void testEliminateBigBusinessesMethod() {
+  public void testFilterBySmallBusinessesMethod() {
     SmallCityService testSmallCityService = new SmallCityService();
-    testSmallCityService.createUserWithGeolocation(testLocation);
```
AbakirH, line 160: Line that was deleted

Issue #35, opened 2020-07-16
Fix asynchronous xmlhttp POST request


LindseyWeiskopf:
Find a way of asynchronously calling xmlhttp without caching user entries or giving inaccurate results

LindseyWeiskopf:
Solutions attempted:

FAILED
- on ready state change event handler to see when xmlhttp POST sent
- set xmlhttp to 'no-cache' and 'no-store'
- promise threading for xmlhttp request and fetchList function call
- create an async function with await calls

WORKED (but deprecated)
- change xmlhttp post request from asynchronous to synchronous

WORKS (currently)
- remove xmlhttp request and send user data as a query string to a getRequest which creates a smallCityService and finds the small businesses synchronously

Method
- logging server output before and after data is transferred from one object to another
- logging to the console before and after GET and POST requests

PR #37, opened 2020-07-16
Fix async by consolidation

LindseyWeiskopf:
Summary: Remove xmlhttp requests and send user location in query string to doGet() function

To remove the asynchronous nature of the code, I removed the xmlHttp request and sent the user's location (zip or coordinate) to the servlet via a fetch request. This way, all of the logic happens sequentially in both the servlet and in the javascript without needing timeouts. I think that this is a good solution because it eliminates any chance is operations happening out of order. 

In the BusinessesService, I also changed the getSmallBusinesses function to return the businesses list to prevent an incomplete list being fetched.

Thank you for joining me on this journey. #35

### DIFF
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -69,7 +69,9 @@ public class SmallCityService {
     queryOfDatabase = businessesService.getBigBusinessFromDatabase();
     businesses = businessesService.removeBigBusinessesFromResults(queryOfDatabase);
   }
-
+  
+  // To be used for unit testing file to be able to get list 
+  // of businesses
   public List<Listing> getBusinesses() {
     return businesses;
   }
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -58,7 +58,6 @@ public class SmallCityService {
   
   public void findAllBusinesses() {
     businesses = businessesService.getBusinessesFromPlacesApi(user);
-
   }
   
   // To be used for unit testing file to be able to 
@@ -83,9 +82,6 @@ public class SmallCityService {
     businessesService = new BusinessesService(businesses);
     findAllBusinesses();
     filterBySmallBusinesses();
-    for (Listing listing : businesses) {
-      LOGGER.info(listing.getName());
-    }
     return businesses;
   }
 }
--- a/src/main/java/com/google/sps/data/SmallCityService.java
+++ b/src/main/java/com/google/sps/data/SmallCityService.java
@@ -43,7 +43,6 @@ public class SmallCityService {
   **/
   public void createUserWithZip(String zipCode) {
     this.user = new User(zipCode);
-    getSmallBusinesses();
   }
   
   /** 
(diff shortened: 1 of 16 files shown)


LindseyWeiskopf:
> Looks like you need to pull the latest changes into this branch and then re-push

I just re-pulled and got the existing map's code. When I merged before with the .env code, I had to revert it to deploy the site and now I'm trying to figure out how to un-revert my revert

src/main/java/com/google/sps/data/SmallCityService.java
```diff
@@ -43,6 +43,7 @@ public SmallCityService() { }
   **/
   public void createUserWithZip(String zipCode) {
     this.user = new User(zipCode);
+    LOGGER.info(this.user.getGeolocation().lat + ", " + this.user.getGeolocation().lng);
```
shaquille-hall, line 46: Leftover bits from debugging? Looks like this can be removed.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,15 +49,12 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
     }
-  }
-
-  @Override
-  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+  
     response.setContentType("application/json;");
     response.setCharacterEncoding("UTF-8");
-    response.getWriter().println(convertToJson(smallCityService.getBusinesses()));
+    response.getWriter().println(convertToJson(smallCityService.getSmallBusinesses()));
```
shaquille-hall, line 57: Can this method be removed altogether? It's only being used in the test now, which should also be updated.

src/main/java/com/google/sps/data/SmallCityService.java
LindseyWeiskopf: Removed it!

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: Yes the method should be removed -- @AbakirH is there a way to do your test without this function?

LindseyWeiskopf: I created an issue to track this #50

AbakirH: I can remove it, but until this pull request is approved.

shaquille-hall: Can we remove it in this pull request?

AbakirH: Never mind I do not think we can remove this function, as the bigbusinessesTest java file, we do not want to test the places API within the first test. Thus, by calling getSmallBusinesses, it overrides my local list and I only want to test the filiterBySmallBusinesses method with the local list in the test. 

A work around to this problem is to make the variable businesses in smallCItyService public so it can be accessed anywhere.

LindseyWeiskopf: Added a comment describing the reason for this function

shaquille-hall: Was this added? I'm not seeing it.

LindseyWeiskopf: It's on line 73-74 of SmallCityService.java

PR #38, opened 2020-07-16
Interactive markers

LindseyWeiskopf:
Summary: Create interactive map feature that highlights listing

I created a marker listener so that when a marker is clicked, the map will zoom to that marker, paginate towards the listing, and highlight the appropriate listing. When the next marker is selected, the past listing should un-highlight 
**Not deployed at the moment

Thank you!

### DIFF
--- a/src/main/webapp/mapScript.js
+++ b/src/main/webapp/mapScript.js
@@ -12,12 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-let highlightedCardId = 0;
+
 
 function initMap(mapLocation) {
   let myLatLng = {lat: mapLocation.lat, lng: mapLocation.lng};
   map = new google.maps.Map(document.getElementById('map'), {
-    zoom: 16,
     center: myLatLng,
     mapTypeControl: false
   });
@@ -27,6 +26,7 @@ function createMarker(listing, cardNumber) {
   let marker = new google.maps.Marker({
     position: {lat: listing.mapLocation.lat, lng: listing.mapLocation.lng},
   });
+  bounds.extend(marker.getPosition());
 
   marker.addListener('click', function() {
     map.setCenter(marker.getPosition());
--- a/src/main/webapp/mapScript.js
+++ b/src/main/webapp/mapScript.js
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+let highlightedId = 0;
+
 function initMap(mapLocation) {
   let myLatLng = {lat: mapLocation.lat, lng: mapLocation.lng};
   map = new google.maps.Map(document.getElementById('map'), {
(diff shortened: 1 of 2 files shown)


src/main/webapp/mapScript.js
```diff
@@ -0,0 +1,44 @@
...
+
+  marker.addListener('click', function() {
+    document.getElementById(highlightedId).style.backgroundColor = '#fff';
+    map.setCenter(marker.getPosition());
+    map.setZoom(16);
+    let firstId = document.getElementById('results-content').firstChild.id;
+    // If card is not on the current page
+    if ((cardNumber-firstId < 0) || (cardNumber-firstId >= 3)) {
+      displayCards(cardNumber-firstId);
+    }
```
shaquille-hall, line 39: Curious. Wouldn't firstId always be 0. So unless cardNumber is also referencing the first card wouldn't this first half effectively just be true. Have you considered using a boolean to track the selected card?

LindseyWeiskopf: Since the results-content section only holds three children, the firstId will be the id of whatever card is first on the current page (ex. I scroll to the third page and then click a marker). The first half would be true only if the card is one of the three on the current page that the user is paginated to. Does that make sense? I decided to not use a boolean because only one card should be selected at a time, and ideally that card will only be unhighlighted when another marker is chosen to avoid confusion. An alternative would be to have the chosen card flash when the marker is clicked if you think that is a better solution then carrying over the highlighted card to the next call of the function.

src/main/webapp/mapScript.js
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+let highlightedId = 0;
```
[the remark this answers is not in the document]
LindseyWeiskopf: Fixed (got rid of the variable all together)

LindseyWeiskopf:
I fixed the page augment, so the markers should not get messed up if the user paginates to a different page. I also changed the highlight so that the card flashes for 5 seconds to avoid carrying over a highlightedCard variable for each click of a marker.

src/main/webapp/mapScript.js
```diff
@@ -12,18 +12,36 @@
...
-function createMarker(listing) {
+function createMarker(listing, cardNumber) {
   let marker = new google.maps.Marker({
     position: {lat: listing.mapLocation.lat, lng: listing.mapLocation.lng},
   });
+  bounds.extend(marker.getPosition());
+
+  marker.addListener('click', function() {
+    map.setCenter(marker.getPosition());
+    map.setZoom(18);
```
shaquille-hall, line 33: Confirming this is meant to be different from the original zoom value. 18 != 13

src/main/webapp/mapScript.js
```diff
@@ -12,18 +12,36 @@
...
+    let firstCardId = document.getElementById('results-content').firstChild.id;
+    // If card is not on the current page
+    if ((cardNumber-firstCardId < 0) || (cardNumber-firstCardId >= 3)) {
+      displayCards(Math.floor((cardNumber-firstCardId)/3)*3);
+    }
+    let currentCard = document.getElementById(cardNumber);
```
shaquille-hall, line 39: nit: Please add a space above this line.

src/main/webapp/mapScript.js
```diff
@@ -12,18 +12,36 @@
...
+    // Cooresponding card flashes for 5 seconds
```
shaquille-hall, line 40: *Corresponding

src/main/webapp/mapScript.js
LindseyWeiskopf: Yes -- @AbakirH found a function where you can fit the bounds of the map to the markers so the initial map should be set to the best fit zoom and then it zooms more when a marker is clicked.

src/main/webapp/mapScript.js
LindseyWeiskopf: fixed (spelling is hard!)

PR #39, opened 2020-07-17
Add big businesses to database

AbakirH:
Summary: Go through a series of checks, to see if a business is a big business and add that to the Google's app engine database.

To check, if a business is a big, first I remove the big businesses from the current list using the already existing database, to stop from adding a duplicate. The program first checks how many locations of a business is in a 50,000 meter radius. If there is more then one business returned using the text search request provided by the places API, I check the number of followers the company has on LinkedIn, if it has more then 50,000 followers, we say it is big business then add it to the database. Then if it does not have enough followers, we do a final check, with the results returned from the text search request to see how many businesses have matching names and different addresses. If there is more then 10  of them, then the company should be a big business and we add it to our database.


src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -110,4 +126,102 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+                                .setApplicationName("linkedinSearch") 
+                                .setGoogleClientRequestInitializer(
+                                      new CustomsearchRequestInitializer(KEY)) 
+                                .build();
+    Customsearch.Cse.List list = cs.cse().list(currentBusinessName).setCx(cx); 
+    List<Result> searchJsonResults = list.execute().getItems();
+    String[] numberOfFollowers;
+    if (searchJsonResults!=null){
+      if(searchJsonResults.size() != 0){
+        Result linkedinBusiness = searchJsonResults.get(0);
```
shaquille-hall, line 169: Is this for one business or all of them?

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -110,4 +126,102 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+        List<Map<String, Object>> resultsMetatags = 
+                        (List) linkedinBusiness.getPagemap().get("metatags");
+        for (Map<String, Object> tag : resultsMetatags) {
+          String b_title = (String) tag.get("og:description");
```
shaquille-hall, line 173: businessTitle?

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -110,4 +126,102 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+          numberOfFollowers = b_title.split(" ");
+          findNumberOfFollowers(numberOfFollowers);
+        }
+      }
+    }
+  }
+
+  private void findNumberOfFollowers(String[] numberOfFollowers) {
+    int companyFollowers = 0;
+    for(int i = 0; i < numberOfFollowers.length; i++) {
```
shaquille-hall, line 183: Just curious, was a forEach loop considered?

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -110,4 +126,102 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+        for (Map<String, Object> tag : resultsMetatags) {
+          String b_title = (String) tag.get("og:description");
+          numberOfFollowers = b_title.split(" ");
+          findNumberOfFollowers(numberOfFollowers);
+        }
+      }
+    }
+  }
+
+  private void findNumberOfFollowers(String[] numberOfFollowers) {
```
shaquille-hall, line 181: To make sure I understand, this is converting the number of followers found on linkedin to an int, then comparing it some minFollowers, right? If that's the case, shouldn't the if statement (:189-193) be inside the for loop? FWIW using this loop also seems like overkill.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -110,4 +126,102 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+      addBigBusinessToDatabase();
+    }else {
+      checkNumberOfSimilarBusinessesInTheArea(currentBusiness.getName());
+    }
+  }
+
+  private void checkNumberOfSimilarBusinessesInTheArea(String businessName){
+    int countNumberOfMatchingBusiness = 0;
+    int i = 0;
+    while(i < similarBusinessesInTheArea.results.length
```
shaquille-hall, line 199: similarBusinessInTheArea is only relevant per business. Can you pass this as a parameter instead of making it global?

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -110,4 +126,102 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+      for(int i = 0; i<allBusinesses.size();i++){
+        currentBusiness = allBusinesses.get(i);
+        TextSearchRequest request = new TextSearchRequest(context)
+                                            .query(currentBusiness.getName())
+                                            .location(latLng)
+                                            .radius(50000);
+        similarBusinessesInTheArea = request.await();
+        if(similarBusinessesInTheArea.results.length > 1){
+          checkBusinessThroughLinkedin(currentBusiness.getName());
+        }
```
LindseyWeiskopf, line 144: Could this also be a forEach loop? Since allBusinesses is a linkedList, the allBusinesses.get(i) function will have to iterate to the ith position of the list for each business. If this is a forEach, you would be able to reference the current business directly.

src/main/java/com/google/sps/data/SmallCityService.java
```diff
@@ -71,6 +71,11 @@ public void filterBySmallBusinesses() {
     queryOfDatabase = businessesService.getBigBusinessFromDatabase();
     businesses = businessesService.removeBigBusinessesFromResults(queryOfDatabase);
   }
+  
+  public void findBigBusinessInList () {
+     businessesService.checkNumberOfLocationsOfBusiness();
+     filterBySmallBusinesses();
```
LindseyWeiskopf, line 77: I think this function may be called twice -- here and on line 88. Is that intentional?

AbakirH: YOU ARE A GENIUS, I fixed this. The program should remove the big business the second that it knows it is one.

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: I made it a while loop instead, so that I can remove the business when I know that it is a big one.

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: I just made it a parameter, I made it global just because it is not used in every function

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Removed this function, found a better way to find the number of followers

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: function does not exist anymore

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Different name now

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: One business only

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -110,4 +126,102 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+                                null) 
+                                .setApplicationName("linkedinSearch") 
+                                .setGoogleClientRequestInitializer(
+                                      new CustomsearchRequestInitializer(KEY)) 
+                                .build();
+    Customsearch.Cse.List list = cs.cse().list(currentBusinessName).setCx(cx); 
+    List<Result> searchJsonResults = list.execute().getItems();
+    String[] numberOfFollowers;
+    if (searchJsonResults!=null){
+      if(searchJsonResults.size() != 0){
```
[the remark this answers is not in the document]
AbakirH: Fixed

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -110,4 +126,102 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+    //Instance Customsearch
+    Customsearch cs = new Customsearch.Builder(
+                                GoogleNetHttpTransport.newTrustedTransport(), 
+                                JacksonFactory.getDefaultInstance(), 
+                                null) 
+                                .setApplicationName("linkedinSearch") 
+                                .setGoogleClientRequestInitializer(
+                                      new CustomsearchRequestInitializer(KEY)) 
+                                .build();
+    Customsearch.Cse.List list = cs.cse().list(currentBusinessName).setCx(cx);
```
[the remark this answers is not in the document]
AbakirH: Done

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -110,4 +126,102 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+
+    }catch(Exception e) {
+      LOGGER.warning(e.getMessage());
+    }  
+  }
+  
+  private void checkBusinessThroughLinkedin(String currentBusinessName) 
+                          throws GeneralSecurityException, IOException {
+    String cx = "001390425498086947771:mhzyevhmmxq"; //Your search engine
+    //Instance Customsearch
```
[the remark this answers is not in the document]
AbakirH: fixed

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -110,4 +126,102 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+      }
+
+    }catch(Exception e) {
+      LOGGER.warning(e.getMessage());
+    }  
+  }
+  
+  private void checkBusinessThroughLinkedin(String currentBusinessName) 
+                          throws GeneralSecurityException, IOException {
+    String cx = "001390425498086947771:mhzyevhmmxq"; //Your search engine
```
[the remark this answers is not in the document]
AbakirH: Just did so

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -110,4 +126,102 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+        similarBusinessesInTheArea = request.await();
+        if(similarBusinessesInTheArea.results.length > 1){
+          checkBusinessThroughLinkedin(currentBusiness.getName());
+        }
+      }
+
+    }catch(Exception e) {
+      LOGGER.warning(e.getMessage());
+    }  
+  }
```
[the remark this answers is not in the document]
AbakirH: I fixed this

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,109 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+    String businessTypes = "BusinessTypes";
+    String address = "Address";
+    String rating = "Rating";
+    String photos = "Photos";
+    Entity businessEntity = new Entity("BigBusinesses");
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    businessEntity.setProperty(title, currentBusiness.getName());
+    businessEntity.setProperty(address, currentBusiness.getFormattedAddress());
+    businessEntity.setProperty(rating, currentBusiness.getRating());
+    businessEntity.setProperty(businessTypes,
```
LindseyWeiskopf, line 232: We definitely want to make sure that the url gets added to the datastore properties -- going to make an issue to track this. It depends which PR gets approved first.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,109 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+    List<Result> searchJsonResults = list.execute().getItems();
+    String[] numberOfFollowers;
+    int companyFollowers = 0;
+    if (searchJsonResults!=null && searchJsonResults.size() != 0) {
+      Result linkedinBusiness = searchJsonResults.get(0);
+      String businessDescription = (String) linkedinBusiness.getPagemap()
+                                                            .get("metatags")
+                                                            .get(0)
+                                                            .get("og:description");
+      if(businessDescription.indexOf(startIndex) != -1
```
LindseyWeiskopf, line 185: nit: A comment here may be helpful to explain what the -1 is doing, etc.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,109 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+                              throws GeneralSecurityException, IOException {
+    String cx = "REDACTED"; 
+    Customsearch cs = new Customsearch.Builder(
+                                GoogleNetHttpTransport.newTrustedTransport(), 
+                                JacksonFactory.getDefaultInstance(), 
+                                null) 
+                                .setApplicationName("linkedinSearch") 
+                                .setGoogleClientRequestInitializer(
+                                      new CustomsearchRequestInitializer(KEY)) 
+                                .build();
```
shaquille-hall, line 173: This one's tricky. See below:

```
 Customsearch cs = new Customsearch.Builder(
       GoogleNetHttpTransport.newTrustedTransport(), JacksonFactory.getDefaultInstance(), null)
            .setApplicationName("linkedinSearch") 
            .setGoogleClientRequestInitializer(new CustomsearchRequestInitializer(KEY)).build();
```

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,109 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+    Customsearch.Cse.List list = cs.cse().list(currentBusinessName).setCx(cx); 
+    List<Result> searchJsonResults = list.execute().getItems();
+    String[] numberOfFollowers;
+    int companyFollowers = 0;
+    if (searchJsonResults!=null && searchJsonResults.size() != 0) {
+      Result linkedinBusiness = searchJsonResults.get(0);
+      String businessDescription = (String) linkedinBusiness.getPagemap()
+                                                            .get("metatags")
+                                                            .get(0)
+                                                            .get("og:description");
```
shaquille-hall, line 184:
```
String businessDescription = 
    (String) linkedinBusiness.getPagemap().get("metatags").get(0).get("og:description");
```

src/main/java/com/google/sps/data/BusinessesService.java
shaquille-hall, line 185: Is "startIndex" defined anywhere? Same as "endIndex" below. These appear to be local variables, but I don't see them anywhere in this function.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -37,6 +48,12 @@
   private final static Logger LOGGER = 
         Logger.getLogger(BusinessesService.class.getName());
   private final int ALLOWED_SEARCH_REQUESTS = 3;
+  private Listing currentBusiness;
+  private Iterator<Listing> businesses;
+  private LatLng latLng;
+  private final int minFollowers = 50000;
+  private final String startIndex = "| ";
+  private final String endIndex = "followers";
```
shaquille-hall, line 56: Constants should be ALL CAPS.

Additionally, neither startIndex nor endIndex are actually indices (integers).

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,109 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+      } else {
+        checkNumberOfSimilarBusinessesInTheArea(currentBusiness.getName(), similarBusinessesInTheArea);
+      }
+    }
+  }
+
+  private void checkNumberOfSimilarBusinessesInTheArea(String businessName, 
+                          PlacesSearchResult[] similarBusinessesInTheArea) {
+    int countNumberOfMatchingBusiness = 0;
+    int i = 0;
```
shaquille-hall, line 204: What is the goal here?

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,109 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+      }
+      i++;
+     }
+     if (countNumberOfMatchingBusiness >= 10) {
+       addBigBusinessToDatabase();
+     }
+   }
+
+  private void addBigBusinessToDatabase(){
+    businesses.remove();
```
shaquille-hall, line 221: Confused. What's being removed here? The entire list of businesses?

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Sounds good

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: This removes the current business that is determined to be a a big business, if you follow the code from the while loop in the checkNumberOfLocationsOfBusiness method. I think you would conclude that the same thing. I can add a comment, or  create a method that makes it more clear. I am doing this, so I do not need to use the removeBigBusinessesFromResults method again, to connect to the database and waste more time.

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Well the i variable allows me to iterate through the results from the textSearchRequest, and countNumberOfMatchingBusiness allows me to check how many businesses within the results have the same corresponding name and different addresses.

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: fixed

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: I fixed the name of the variable, please tell me if you would like me to change it again

AbakirH: I think the best way to communicate what the indexes are for is through a comment right?

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: fixed

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: I followed this as much as I thought I could

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,109 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+        }
+      }
+    } catch(GeneralSecurityException e) {
+      LOGGER.warning(e.getMessage());
+    } catch(IOException e) {
+      LOGGER.warning(e.getMessage());
+    } catch(ApiException e) {
+      LOGGER.warning(e.getMessage());
+    } catch(InterruptedException e) {
+      LOGGER.warning(e.getMessage());
```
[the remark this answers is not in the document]
AbakirH: 1. So I moved the try and catch for only the await method call. However, I am not sure what is throwing the GeneralSecurityException, I can not find it online, but it might be the documenetaion is outdated a bit, because if I remove this the program tell me I need it. 
2. Thank you for this suggestion

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Not sure if I should as, indexOf returns -1 instead of null if something is not found in the string. I think this is something someone can easily figure out or it is already known

shaquille-hall: Code should first be as readable as possible, giving all of the necessary context using just the language constructs. This is where meaningful naming conventions become crucial. Comments are secondary to this. They should be used to value that other naming constructs can't without becoming unwieldy.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,105 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+      }
+    }
+  }
+
+  private void checkNumberOfSimilarBusinessesInTheArea(String businessName, 
+                          PlacesSearchResult[] similarBusinessesInTheArea) {
+    int countNumberOfMatchingBusiness = 0;
+    int i = 0;
+    while (i < similarBusinessesInTheArea.length 
+          && countNumberOfMatchingBusiness < 10) {
```
shaquille-hall, line 202: Please make 10 a constant, and you might want to decrease it. How was it settled on?

src/main/java/com/google/sps/data/BusinessesService.java
shaquille-hall: This seems misplaced. The method is should focus on adding entries to the database. Please move it to the point where this method is called. I've also noted that this variable can be removed. Please see my previous comments.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -72,7 +89,7 @@ public PreparedQuery getBigBusinessFromDatabase(){
   }
   
   public List<Listing> getBusinessesFromPlacesApi(MapLocation mapLocation) {
-    LatLng latLng = 
+    latLng =
```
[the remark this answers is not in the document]
AbakirH: I made this change, as the Text Search Request is very similar to the Nearby Search Request and both need the location of the person at the time of use. If we make this variable global, we can use this throughout different functions where we use the Places API.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,105 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
-}
\ No newline at end of file
+
+  public void checkNumberOfLocationsOfBusiness() {
+    GeoApiContext context = new GeoApiContext.Builder()
+            .apiKey(KEY)
+            .build();
+    businesses =  allBusinesses.iterator(); 
+    
+    while (businesses.hasNext()) {
```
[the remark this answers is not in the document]
AbakirH: The problem with doing a for each loop with the linked list, is if I try to remove a an element from the list then I will get the ConcurrentModificationException error. To not get this error I have to use a while loop, using an iterator with it. Also, I have to call .next() before I can use the .remove method for an iterator. This can be found here java67.com/2018/12/how-to-remove-objects-or-elements-while-iterating-Arraylist-java.html.

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Got it

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,105 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+      } else {
+        checkNumberOfSimilarBusinessesInTheArea(currentBusiness.getName(),
+                                                similarBusinessesInTheArea);
+      }
+    }
+  }
+
+  private void checkNumberOfSimilarBusinessesInTheArea(String businessName, 
+                          PlacesSearchResult[] similarBusinessesInTheArea) {
+    int countNumberOfMatchingBusiness = 0;
```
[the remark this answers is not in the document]
AbakirH: fixed

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,105 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+      String businessDescription = 
+        (String) linkedinBusiness.getPagemap().get("metatags").get(0).get("og:description");
+      if(businessDescription.indexOf(START_SUBSTRING) != -1 
+          && businessDescription.indexOf(END_SUBSTRING) != -1){
+        String followers = businessDescription.substring(
+                                businessDescription.indexOf(START_SUBSTRING) + 2, 
+                                businessDescription.indexOf(END_SUBSTRING) - 1);
+        try{
+          companyFollowers = 
+                    Integer.parseInt(followers.replaceAll(",", ""));
```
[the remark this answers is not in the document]
AbakirH: fixed

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -37,6 +48,12 @@
   private final static Logger LOGGER = 
         Logger.getLogger(BusinessesService.class.getName());
   private final int ALLOWED_SEARCH_REQUESTS = 3;
+  private Listing currentBusiness;
+  private Iterator<Listing> businesses;
+  private LatLng latLng;
+  private final int MINFOLLOWERS = 50000;
+  private final String START_SUBSTRING = "| ";
+  private final String END_SUBSTRING = "followers";
```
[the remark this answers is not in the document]
AbakirH: fixed

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -54,7 +71,7 @@ public PreparedQuery getBigBusinessFromDatabase(){
   }
 
   public List<Listing> removeBigBusinessesFromResults(PreparedQuery queryOfDatabase){
-    Iterator<Listing> businesses =  allBusinesses.iterator();
+    businesses =  allBusinesses.iterator();
```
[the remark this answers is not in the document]
AbakirH: I can not call allBusinesses.iterator().next() within a while loop, as the program will not go to the next element of the list. Instead as seen in all these examples: https://www.geeksforgeeks.org/how-to-use-iterator-in-java/. This is because, every time you call .iterator method, you are creating a new iterator, which results in only getting the first element of the list.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,105 @@ private void addListingToBusinesses(PlacesSearchResult place) {
     allBusinesses.add(new Listing(name, formattedAddress, 
           placeLocation, rating, photos, types));
   }
-}
\ No newline at end of file
+
+  public void checkNumberOfLocationsOfBusiness() {
+    GeoApiContext context = new GeoApiContext.Builder()
+            .apiKey(KEY)
+            .build();
```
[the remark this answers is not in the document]
AbakirH: Fixed

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,105 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+    businesses =  allBusinesses.iterator(); 
+    
+    while (businesses.hasNext()) {
+      currentBusiness = businesses.next();
+      TextSearchRequest request = new TextSearchRequest(context)
+                                          .query(currentBusiness.getName())
+                                          .location(latLng)
+                                          .radius(50000);
```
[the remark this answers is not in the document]
AbakirH: fixed

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,105 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+      try {                                       
+        PlacesSearchResult[] similarBusinessesInTheArea = request
+                                                            .await()
+                                                            .results;
```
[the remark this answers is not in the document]
AbakirH: fixed

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,105 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+        JacksonFactory.getDefaultInstance(), null) 
+              .setApplicationName("linkedinSearch") 
+              .setGoogleClientRequestInitializer(new CustomsearchRequestInitializer(KEY)) 
+              .build();
+
+    Customsearch.Cse.List list = cs.cse().list(currentBusinessName).setCx(cx); 
+    List<Result> searchJsonResults = list.execute().getItems();
+    String[] numberOfFollowers;
+    int companyFollowers = 0;
+    if (searchJsonResults!=null && searchJsonResults.size() != 0) {
```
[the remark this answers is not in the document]
AbakirH: fixed

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -111,4 +128,105 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+      Result linkedinBusiness = searchJsonResults.get(0);
+      String businessDescription = 
+        (String) linkedinBusiness.getPagemap().get("metatags").get(0).get("og:description");
+      if(businessDescription.indexOf(START_SUBSTRING) != -1 
+          && businessDescription.indexOf(END_SUBSTRING) != -1){
```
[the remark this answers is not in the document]
AbakirH: Thank you I changed both checks to use .contains instead

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Got it, I am now using the .contains method, so the code should be more readable now.

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Yes I fixed this issue

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: I changed the number to 5, as it should hopefully get all the big businesses in the area

src/main/java/com/google/sps/data/BusinessesService.java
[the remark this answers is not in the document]
shaquille-hall: What error are you getting when you try? According to the doc: https://docs.oracle.com/javase/8/docs/api/java/util/List.html this is supported. Even so, if you're only iterating through the list, a foreach loop will give you the same result with less boilerplate.

src/main/java/com/google/sps/data/BusinessesService.java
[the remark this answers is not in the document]
shaquille-hall: Ok. So it's more than just iteration. In that case you can do a traditional for loop, something like:

```
totalNumberOfBusinesses = allBusinesses.size();
for (for i = 0; i < totalNumberOfBusiness; i++) {
    ...
    if (businessToRemove == true) {
        allBusinesses.remove(i);
    }
    ...
}

File: src/main/java/com/google/sps/data/BusinessesService.java (hunk 32 above)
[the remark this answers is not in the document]
Comment (shaquille-hall) [reply src/main/java/com/google/sps/data/BusinessesService.java]: "== true", to me, is a lot more intuitive than "!= false".

File: src/main/java/com/google/sps/data/BusinessesService.java [hunk 33]
```diff
@@ -31,11 +44,19 @@
...
-  private List<Listing> allBusinesses;
   private final String KEY = "REDACTED";
   private final static Logger LOGGER = 
         Logger.getLogger(BusinessesService.class.getName());
   private final int ALLOWED_SEARCH_REQUESTS = 3;
+  private final int MINFOLLOWERS = 50000;
+  private final int SMALL_BUSINESSES_DISPLAYED = 15;
+  private final String START_SUBSTRING = "| ";
+  private final String END_SUBSTRING = "followers";
+  private final int MIN_NUMBER_OF_MATCHING_BUSINESSES = 5;
```
shaquille-hall, line 55: Would "ALLOWED_NUMBER_OF_MATCHING_BUSINESSES" be better here, or "MAX_..." based on what's happening this is bing used as an upper bound to check against.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -107,4 +128,104 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+        String followers = businessDescription.substring(
+                                businessDescription.indexOf(START_SUBSTRING) + 2, 
+                                businessDescription.indexOf(END_SUBSTRING) - 1);
+        try{
+          companyFollowers = Integer.parseInt(followers.replaceAll(",", ""));
+        } catch (NumberFormatException e){
+            LOGGER.warning(e.getMessage());
+        }
+      }
+      if (companyFollowers > MINFOLLOWERS) {
```
shaquille-hall, line 187: *MIN_FOLLOWERS

src/main/java/com/google/sps/data/BusinessesService.java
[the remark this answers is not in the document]
AbakirH: I keep getting this error (java.util.ConcurrentModificationException), when I try to remove a element from a list inside a for loop or while loop and the only way around that was using an iterator. Instead of removing the big business when it is found, I will recall the remove filterBySmallBusinesses method inside the SmallCityService class after all the big businesses have been added.

src/main/java/com/google/sps/data/BusinessesService.java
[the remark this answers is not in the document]
AbakirH: Will do

src/main/java/com/google/sps/data/BusinessesService.java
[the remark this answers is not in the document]
AbakirH: fixed

src/main/java/com/google/sps/data/BusinessesService.java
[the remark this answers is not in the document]
AbakirH: If you look at this comment I just made, it has never worked for me. It does not seem you can't remove an element within a loop. https://github.com/googleinterns/SmallCity/pull/39#discussion_r461176814

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: I decided to use ALLOWED_NUMBER_OF_MATCHING_BUSINESSES

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Fixed

src/main/java/com/google/sps/data/BusinessesService.java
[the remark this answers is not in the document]
AbakirH: Also I stop the while loop after fifteen small business have been found, as that is the only thing that displays on the screen.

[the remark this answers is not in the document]
shaquille-hall: Further thought. I think you're right. The snippet above is invalid.

src/main/java/com/google/sps/data/BusinessesService.java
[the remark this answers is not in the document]
shaquille-hall: See my comment below on this.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -108,4 +129,104 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
   }
-}
\ No newline at end of file
+
+  public void checkNumberOfLocationsOfBusiness() {
+    GeoApiContext context = 
+      new GeoApiContext.Builder().apiKey(KEY).build();
+    numberOfSmallBusinesses = 0;
+    Iterator<Listing> businesses =  allBusinesses.iterator();
+    while(businesses.hasNext()
```
shaquille-hall, line 138: nit: Add a space above this line to make this more readable.

Thanks for sticking through this.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -108,4 +129,104 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+          && numberOfSmallBusinesses < SMALL_BUSINESSES_DISPLAYED) {
+      Listing currentBusiness = businesses.next();
+      TextSearchRequest request = 
+        new TextSearchRequest(context).query(currentBusiness.getName())
+          .location(latLng).radius(50000);
+      try {
```
shaquille-hall, line 144: nit: Add a space above this line to make this more readable.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -108,4 +129,104 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+    while(businesses.hasNext()
+          && numberOfSmallBusinesses < SMALL_BUSINESSES_DISPLAYED) {
+      Listing currentBusiness = businesses.next();
+      TextSearchRequest request = 
+        new TextSearchRequest(context).query(currentBusiness.getName())
+          .location(latLng).radius(50000);
+      try {                                       
+        PlacesSearchResult[] similarBusinessesInTheArea = 
+          request.await().results;
+        if (similarBusinessesInTheArea.length > 1){
```
shaquille-hall, line 147: nit: Add a comment above this about why the LinkedIn check is useful.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -108,4 +129,104 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+        JacksonFactory.getDefaultInstance(), null) 
+              .setApplicationName("linkedinSearch") 
+              .setGoogleClientRequestInitializer(new CustomsearchRequestInitializer(KEY)) 
+              .build();
+
+    Customsearch.Cse.List list = cs.cse().list(currentBusiness.getName()).setCx(cx); 
+    List<Result> searchJsonResults = list.execute().getItems();
+    String[] numberOfFollowers;
+    int companyFollowers = 0;
+    if (searchJsonResults != null && searchJsonResults.size() != 0) {
```
shaquille-hall, line 173: nit: Add a space above this line to make this more readable.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -108,4 +129,104 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+      Result linkedinBusiness = searchJsonResults.get(0);
+      String businessDescription = 
+        (String) linkedinBusiness.getPagemap().get("metatags").get(0).get("og:description");
+      if(businessDescription.contains(START_SUBSTRING) == true
```
shaquille-hall, line 177: nit:  Add a space above this line to make this more readable.

Also, you can remove the "== true" part if you want. The evaluation will still work. Apologies for missing that in my last round.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -108,4 +129,104 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+    if (searchJsonResults != null && searchJsonResults.size() != 0) {
+      Result linkedinBusiness = searchJsonResults.get(0);
+      String businessDescription = 
+        (String) linkedinBusiness.getPagemap().get("metatags").get(0).get("og:description");
+      if(businessDescription.contains(START_SUBSTRING) == true 
+          && businessDescription.contains(END_SUBSTRING) == true){
+        String followers = businessDescription.substring(
+                                businessDescription.indexOf(START_SUBSTRING) + 2, 
+                                businessDescription.indexOf(END_SUBSTRING) - 1);
+        try{
```
shaquille-hall, line 182: nit: Add a space above this line to make this more readable.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -108,4 +129,104 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+          companyFollowers = Integer.parseInt(followers.replaceAll(",", ""));
+        } catch (NumberFormatException e){
+            LOGGER.warning(e.getMessage());
+        }
+      }
+      if (companyFollowers > MIN_FOLLOWERS) {
```
shaquille-hall, line 188: nit: Add a space above this line to make this more readable.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -108,4 +129,104 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+                                                similarBusinessesInTheArea);
+      }
+    }
+  }
+
+  private void checkNumberOfSimilarBusinessesInTheArea(Listing currentBusiness, 
+                          PlacesSearchResult[] similarBusinessesInTheArea) {
+    int numberOfMatchingBusinesses = 0;
+    int i = 0;
+    while (i < similarBusinessesInTheArea.length
```
shaquille-hall, line 201: nit: Add a space above this line to make this more readable.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -108,4 +129,104 @@ private void addListingToBusinesses(PlacesSearchResult place) {
...
+    }
+  }
+
+  private void checkNumberOfSimilarBusinessesInTheArea(Listing currentBusiness, 
+                          PlacesSearchResult[] similarBusinessesInTheArea) {
+    int numberOfMatchingBusinesses = 0;
+    int i = 0;
+    while (i < similarBusinessesInTheArea.length 
+          && numberOfMatchingBusinesses < ALLOWED_NUMBER_OF_MATCHING_BUSINESSES) {
+        if (similarBusinessesInTheArea[i].name.contains(currentBusiness.getName())
```
shaquille-hall, line 203: Just curious. Should this be contains or equals? I'm sure we can find cases that void either side, but it might be something you want to think about.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -108,4 +129,104 @@ private void addListingToBusinesses(PlacesSearchResult place) {
     allBusinesses.add(new Listing(name, formattedAddress, 
           placeLocation, rating, photos, types));
   }
-}
\ No newline at end of file
+
+  public void checkNumberOfLocationsOfBusiness() {
```
shaquille-hall, line 133: This method initiates a few different checks to verify a big business. Please enumerate those with a comment here. It may also be worth it to change the name of this method appropriately. It's doing more than checking the number of locations.

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: I changed it to use the equals method for the time being, because on further thought it the results that are returned from the text search request should have the same name as the results are coming from the nearby search request results. I was scared that the name might differ a bit and would result in the business not being caught as a big. Nevertheless, I think it is the smart thing to do, unless I notice something later on.

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: i changed the name of the function.

PR #40, opened 2020-07-17
Refactor User logic to UserService and alter tests

LindseyWeiskopf:
Summary: Refactor most of User class logic into a UserService class

In this PR, I moved the zipCodeToMapLocation and User constructors into a UserService class to remove some of the load from the User object. In addition, anywhere that there was a user instantiation, I changed it to a userService object. I also converted the UserTest class into a UserService test class since the logic being tested was also moved. The getBusinessesfromPlacesAPI function now takes a mapLocation as input as opposed to a User instance to remove unnecessary information from being passed into the function.

Thanks!

### DIFF
--- a/src/main/java/com/google/sps/data/UserService.java
+++ b/src/main/java/com/google/sps/data/UserService.java
@@ -1,0 +1,54 @@
+package com.google.sps.data;
+
+import com.google.maps.GeocodingApi;
+import com.google.maps.model.AddressComponentType;
+import com.google.maps.model.AddressType;
+import com.google.maps.model.ComponentFilter;
+import com.google.maps.model.GeocodingResult;
+import com.google.maps.GeoApiContext;
+import com.google.maps.model.LatLng;
+import com.google.maps.model.LocationType;
+import java.util.Arrays;
+import java.util.logging.Logger;
+import io.github.cdimascio.dotenv.Dotenv;
+
+public class UserService {
+  
+  private Dotenv dotenv = Dotenv.configure().load();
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/sps/data/UserService.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.maps.GeoApiContext;
+import com.google.maps.model.LatLng;
+import com.google.maps.model.LocationType;
+import java.util.Arrays;
+import java.util.logging.Logger;
+import io.github.cdimascio.dotenv.Dotenv;
+
+public class UserService {
+  
+  private Dotenv dotenv = Dotenv.configure().load();
```
ethanimooney, line 17: You left the dotenv in here which is fine but if we don't end up using it let's make sure to remove, do we already have an issue tracking our progress on the API keys?

shaquille-hall: @AbakirH

LindseyWeiskopf:
I removed the double constructor in the UserService class and created functions that have the proper parameter. Let me know if this is a sound way to implement this.

src/main/java/com/google/sps/data/UserService.java
shaquille-hall: Per our conversation in standup, let's delete the Dotenv setup. @AbakirH will handle this separately.

LindseyWeiskopf: The dotenv setup is now deleted

PR #59, opened 2020-07-22
Change Image Loading, Other Responsive Changes

ethanimooney:
TL;DR : Changed the way the images are loaded in, reducing page load time, other responsive bugs fixed
LINK: www.step2020-smallcity.appspot.com/

**NOTE** : changes may not be visible at the link, as abakir@ is deploying in order to test things

**mapScript.js** : Changed the card highlight behavior when a marker is clicked from full background highlight to border hightlight

**paginationScript.js** : Added code to change the loading time for business images. Now, as a card is about to be paginated to, the script checks the photoReferencesArray to see if there is a photo for the current business, if so, it calls loadImage, passing the place (results-image-element) to load the image to and the photoReference of the image. loadImage then sets the source of the image using the given format.

**script.js** : Added the photoReferencesArray, which stores the first photo reference for each business returned with the fetch. If the business has no photo, it stores 'none', in order to indicate to paginationScript that there is no image, and that it should not try to load one.

**style.css** : Various responsiveness changes, notably image sizing should now be consistent, and the cards should fit inside of their container without the need to scroll anymore. There are other minor issues with responsiveness that I am still working on, but for now this is pretty solid.

### DIFF
--- a/src/main/webapp/paginationScript.js
+++ b/src/main/webapp/paginationScript.js
@@ -21,6 +21,8 @@ const resultsChildren = resultsContent.childNodes;
 const TOTAL_CARDS_TO_DISPLAY = 3;
 const MAX_LIST_VIEW_NUMBER = 15;
 
+
+
 //Display the initial 3 cards in the list
 function initialDisplay() {
   resultsContent.innerHTML = '';
@@ -57,7 +59,10 @@ function displayCards(listAugment) {
     resultsContent.innerHTML = '';
     for (let i = currentFirstCardIndex; i < (currentFirstCardIndex + TOTAL_CARDS_TO_DISPLAY); i++) {
       if (photoReferencesArray[i] != 'none') {
-        loadImage(listingsArray[i].children[0].children[0], photoReferencesArray[i]);
+        //The actual image element to which the image src will be applied
+        let imageElement = listingsArray[i].children[0].children[0];
+        
+        loadImage(imageElement, photoReferencesArray[i]);
       }
 
       resultsContent.appendChild(listingsArray[i]);
--- a/src/main/webapp/paginationScript.js
+++ b/src/main/webapp/paginationScript.js
@@ -56,7 +56,19 @@ function displayCards(listAugment) {
 
     resultsContent.innerHTML = '';
     for (let i = currentFirstCardIndex; i < (currentFirstCardIndex + TOTAL_CARDS_TO_DISPLAY); i++) {
+      if (photoReferencesArray[i] != 'none') {
+        loadImage(listingsArray[i].children[0].children[0], photoReferencesArray[i]);
+      }
+
       resultsContent.appendChild(listingsArray[i]);
     }
   }
 }  
+
+function loadImage(listingImage, photoReference) {
+  const KEY = 'AIzaSyDDIsG-SJAZ69ZoOecmfbXOB7ZIS4pZkAw';
+  let maxwidth = 400;
+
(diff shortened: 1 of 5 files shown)


ethanimooney:
I feel like I explained my photo loading solution poorly, so here is another:

From the fetch, when createResultCard() is called, it passes an array **photos[]**, of photo references. In createResultCard(), it ensures that there are photo references passed (as some businesses will not have any), and if there are, it takes the first photo reference (photos[0]) and adds it to photoReferencesArray in the index corresponding to the card being created. If there is no photo references for a certain business, the function will append 'none' in it's place.

After this, the photos are not touched until the paginationScript, when displayCards() is called. In this case, in the for loop that appends the cards about to be displayed to resultsContent, right before it appends a card, it checks to see if there is a photoReference in photoReferencesArray for the card, by checking if the photoReferencesArray index != 'none'. 

If there is a photo reference, it calls loadImage(), passing in the image element for the card, and the photoReference itself. loadImage() then sets the image element's source attribute using the provided url structure. At this point, a call is made to maps and the photo is displayed, magic!

src/main/webapp/paginationScript.js
```diff
@@ -56,7 +56,19 @@ function displayCards(listAugment) {
 
     resultsContent.innerHTML = '';
     for (let i = currentFirstCardIndex; i < (currentFirstCardIndex + TOTAL_CARDS_TO_DISPLAY); i++) {
+      if (photoReferencesArray[i] != 'none') {
+        loadImage(listingsArray[i].children[0].children[0], photoReferencesArray[i]);
```
LindseyWeiskopf, line 60: nit: a comment here explaining what the .children[0].children[0] is representing may be helpful

src/main/webapp/paginationScript.js
```diff
@@ -56,7 +56,19 @@ function displayCards(listAugment) {
...
+      }
+
       resultsContent.appendChild(listingsArray[i]);
     }
   }
-}  
\ No newline at end of file
+}  
+
+function loadImage(listingImage, photoReference) {
```
LindseyWeiskopf, line 68: Have you thought about adding a placeholder image for the listings that don't have a photoreference?

src/main/webapp/paginationScript.js
ethanimooney: Better yet, I added a variable imageElement, with a comment to explain further

src/main/webapp/paginationScript.js
ethanimooney: Yes, working on this just didn't get it added in this pr

src/main/webapp/script.js
```diff
@@ -44,6 +44,7 @@ function getZipCode() {
 
 //Array of the (currently 6 for this demo build) 15 listings gathered from the fetch request
 let listingsArray = [];
+let photoReferencesArray = [];
```
shaquille-hall, line 47: Instead of two parallel arrays, could this have been done with one. That way each element in the listingsArray would 'have' a photoreference?

ethanimooney:
I was able to combine the two arrays by creating a resultsCardObject object that contains the resultCard and the photoReference. I renamed the combined arrays to resultsCardsArray[], to provide more clarity.

I also changed the paginationScript a bit by creating a variable cardToAppend, which is the element in resultsCardsArray about to get appended to the results-content div. This just makes the script a bit cleaner

src/main/webapp/script.js
ethanimooney: Done

Issue #43, opened 2020-07-17
Create loading bar


LindseyWeiskopf:
Create loading bar for long fetch of data

PR #63, opened 2020-07-22
Loader implementation

LindseyWeiskopf:
Summary: Create loader circle for delays in fetch of listings

index.html: Added a loader div which contains the rotating circle and a loader-container div that is a partially transparent dark grey rectangle that overlays on top of the whole page. 

script.js: Added the initiateLoader() function that changes the display of the loader divs from 'none' to 'block' and changes the opacity of the map div to also look faded. I also added a removeLoader() function that restores the display.

style.css: Styling for the loader circle and loader-container

[Image of loader in action](https://screenshot.googleplex.com/rb9z3svnQEK)

#43

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -43,6 +43,9 @@ to represent the company, product, or service to which they refer. -->
     <title>smallcity</title>
   </head>
   <body>
+    <div id="loader-container">
+      <div id="loader"></div>
+    </div>
     <header>
       <h1 class="logo">smallcity</h1>
       <button class="geolocate-button" onclick="getGeolocation()"><i class="fas fa-location-arrow"></i> locate me</button>
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -43,8 +43,8 @@ to represent the company, product, or service to which they refer. -->
     <title>smallcity</title>
   </head>
   <body>
-    <div id="loader-container">
-      <div id="loader"></div>
+    <div id="loader-circle-container">
+      <div id="loader-circle"></div>
     </div>
     <header>
       <h1 class="logo">smallcity</h1>
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -30,6 +30,7 @@ function displayLocation(position) {
   let lat = position.coords.latitude;
   let lng = position.coords.longitude;
   fetchList('/data?lat=' + lat + "&lng=" + lng);
+  initiateLoader();
 }
 
 function displayError() {
@@ -40,6 +41,13 @@ function displayError() {
 function getZipCode() {
   let zip = document.getElementById('zipCode').value;
   fetchList('/data?zipCode=' + zip);
+  initiateLoader();
+}
+
+function initiateLoader() {
+  document.getElementById('loader').style.display = 'block';
+  document.getElementById('loader-container').style.display = 'block';
+  document.getElementById('map').style.opacity = '0.5';
 }
 
 //Array of the (currently 6 for this demo build) 15 listings gathered from the fetch request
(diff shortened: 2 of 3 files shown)


src/main/webapp/style.css
```diff
@@ -111,10 +111,42 @@ Main Content Container Styling
...
+  width: 200px;
+  height: 200px;
+  animation: spin 2s linear infinite;
+  display: none;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  opacity: 1;
+}
+
```
ethanimooney, line 144: nit: According to the style guides, the individual attributes here need to be in alphabetical order

src/main/webapp/script.js
```diff
@@ -61,9 +69,16 @@ function fetchList(queryString) {
...
+    map.fitBounds(bounds);  
+    removeLoader();
   });
 }
+ 
+function removeLoader() {
+  document.getElementById('loader').style.display = 'none';
+  document.getElementById('loader-container').style.display = 'none';
+  document.getElementById('map').style.opacity = '1';
+}
```
ethanimooney, line 81: nit: I might move the initiateLoader and removeLoader declarations to be one after the other, just for clarity's sake

src/main/webapp/script.js
```diff
@@ -40,6 +41,13 @@ function displayError() {
 function getZipCode() {
   let zip = document.getElementById('zipCode').value;
   fetchList('/data?zipCode=' + zip);
+  initiateLoader();
+}
+
+function initiateLoader() {
+  document.getElementById('loader').style.display = 'block';
+  document.getElementById('loader-container').style.display = 'block';
```
ethanimooney, line 49: This might be unnecessary, but you could create global variables for loader and loader-container, and just call the .style.display on those. Might make the individual functions a bit clearer and avoid having to use the getElementById function each time.

src/main/webapp/script.js
```diff
@@ -40,6 +41,13 @@ function displayError() {
...
+  document.getElementById('map').style.opacity = '0.5';
```
ethanimooney, line 50: Can do the same for the map as well, forgot to mention

src/main/webapp/style.css
```diff
@@ -111,10 +111,42 @@ Main Content Container Styling
...
+@keyframes spin {
```
ethanimooney, line 145: nit: I might add a comment explaining what this is for

src/main/webapp/script.js
```diff
@@ -61,9 +69,16 @@ function fetchList(queryString) {
...
-    map.fitBounds(bounds);
+    map.fitBounds(bounds);  
+    removeLoader();
   });
 }
+ 
+function removeLoader() {
+  document.getElementById('loader').style.display = 'none';
+  document.getElementById('loader-container').style.display = 'none';
+  document.getElementById('map').style.opacity = '1';
```
shaquille-hall: Can you add appropriate classes to the css file, and assign those here instead? Same as above.

src/main/webapp/style.css
LindseyWeiskopf: it is now alphabetized!

src/main/webapp/script.js
LindseyWeiskopf: Done

src/main/webapp/style.css
LindseyWeiskopf: You got it!

src/main/webapp/script.js
LindseyWeiskopf: Just created the classes, let me know if the change makes sense.

src/main/webapp/index.html
```diff
@@ -43,6 +43,9 @@
     <title>smallcity</title>
   </head>
   <body>
+    <div id="loader-container">
+      <div id="loader"></div>
```
shaquille-hall, line 47: Caveat: this is more of a preference, than anything else. Reading this, and the method like "initiateLoader" I'd expect code for some process that actually loads or kickstarts something. This is is a ui indicator that doesn't do this. If everyone on your team is ok with it, it's fine. If not, perhaps a name that indicates it's more of a progress bar ordeal. Your comment below calls this a 'loader circle'. I find that a bit clearer.

LindseyWeiskopf: I changed all of the names to loader-circle or loaderCircle

PR #64, opened 2020-07-22
Add links to website buttons

LindseyWeiskopf:
Summary: Implement Place Details request to get the website of the listings and link to listings on page cards

Listing.java: Add url parameter and getUrl() function

BusinessesService: Implement the Place Details Request of the places API with the place Ids from the Nearby Search Request to get the website for each listing. If the listing does not have a website listed on their google profile, the function will return a link to all of the information that google has about that place (essentially Google Maps landing page for business). The website is added to the listing object for each business.

script.js: using the infrastructure that @ethanimooney  built, I added a click listen to the 'Visit Website' buttons that redirects the user to a new window with the listing's website.

BigBusinessTest.java: I added a website to all of the example listings.

NOTE: once this is deployed, we may have to weigh the pro's and con's of this feature depending on the runtime that this adds.


src/main/webapp/script.js
```diff
@@ -108,7 +109,9 @@ function createResultCard(name, address, photos, rating, totalCardCount) {
   const websiteButton = document.createElement('button');
   websiteButton.className = 'results-website-button';
   websiteButton.innerText = 'Visit Website';
-  // TODO: Add website link
+  websiteButton.addEventListener('click', function() {
+    window.open(website);
+  });
```
ethanimooney, line 114: nit: I might add a comment here noting that this is equivalent to onclick in html, just to provide a bit more clarity.

src/main/webapp/script.js
```diff
@@ -108,7 +109,9 @@ function createResultCard(name, address, photos, rating, totalCardCount) {
   const websiteButton = document.createElement('button');
   websiteButton.className = 'results-website-button';
   websiteButton.innerText = 'Visit Website';
```
ethanimooney, line 111: I would do this but you'll be able to much quicker since you already have this PR out. Let's make it so if the url passed is not for their personal website, the button text reads "Visit on Google Maps".

This should be as simple as adding an if statement to check whether the url returned starts with 'maps.google.com' if so, websiteButton.innerText = 'Visit on Google Maps" , otherwise, .....etc.

Let me know if you have any questions about this, I probably explained badly

shaquille-hall: Sounds like a good idea to me.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -98,8 +101,25 @@ public PreparedQuery getBigBusinessFromDatabase(){
...
+      if (result.website != null) {
+        return (result.website.toString());
+      }
+      else {
+        return (result.url.toString());
+      }
+    } catch(Exception e) {
+      LOGGER.warning(e.getMessage());
+    }
+    return "NO URL";
```
shaquille-hall, line 119: Hmm. What is the expected flow for this failure case? This is still returning a string, just as if everything is ok, and the call to obtain a url succeeded. How is this being handled?

src/main/webapp/script.js
LindseyWeiskopf: Done

src/main/webapp/script.js
LindseyWeiskopf: Just made this change and also created a different button message for when the website it unavailable

src/main/java/com/google/sps/data/BusinessesService.java
LindseyWeiskopf: The only way that this return value would be reached is if the try fails, so the warning message would be sent from the error. I changed the return value to an empty string and on the front end, if the website is empty, there will be a website unavailable button message. Is that better for the failure case?

src/main/webapp/script.js
```diff
@@ -107,8 +108,19 @@ function createResultCard(name, address, photos, rating, totalCardCount) {
 
   const websiteButton = document.createElement('button');
   websiteButton.className = 'results-website-button';
-  websiteButton.innerText = 'Visit Website';
-  // TODO: Add website link
+  if (website.includes('maps.google.com')) {
+    websiteButton.innerText = 'Visit Website on Google Maps';
```
ethanimooney, line 112: I would make this "Visit Location on Google Maps"

src/main/webapp/script.js
```diff
@@ -107,8 +108,19 @@ function createResultCard(name, address, photos, rating, totalCardCount) {
...
+    linkWebsite(website, websiteButton);
+  }
+  else if (website === '') {
```
ethanimooney, line 115: This might be a case we need to address further, I'm not sure having just a gray rectangle is the best solution but this works for now as this will take me doing some frontend-magic to make it change. 

Will make an issue to track, though will this case be very prevalent as all of our results are coming from Google Maps and theoretically should have a listing?

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -98,8 +101,26 @@ public PreparedQuery getBigBusinessFromDatabase(){
     }  
     return allBusinesses;
   }
+  
+  private String getUrlFromPlaceDetails(GeoApiContext context, String placeId) {
+    try {
+      PlaceDetails result = new PlaceDetailsRequest(context)
+            .placeId(placeId)
+            .await();
```
shaquille-hall, line 109:
```
 PlaceDetails result = 
    new PlaceDetailsRequest(context).placeId(placeId).await();
```

src/main/webapp/script.js
LindseyWeiskopf: Changed!

src/main/webapp/script.js
LindseyWeiskopf: Theoretically, this case should never happen since the places API should always at least return their reference on maps. I could only foresee this happening if the entire API call fails, and then all of the websites would be unavailable.

src/main/java/com/google/sps/data/BusinessesService.java
LindseyWeiskopf: Fixed

src/main/webapp/script.js
```diff
@@ -73,7 +74,7 @@ function fetchList(queryString) {
  * @param {string} website The url of the business' website
  * @param {int} totalCardCount The number of businesses in the list, used to set a specific id to each card
  */
-function createResultCard(name, address, photos, rating, totalCardCount) {
+function createResultCard(name, address, photos, rating, website, totalCardCount) {
```
shaquille-hall, line 77: nit: Please prefer something like websiteUrl to be more specific.

src/main/webapp/script.js
shaquille-hall: Just chiming in to commend the good thought process here. That being said, I think this workaround also works fine.

src/main/webapp/script.js
```diff
@@ -107,8 +108,19 @@ function createResultCard(name, address, photos, rating, totalCardCount) {
...
   const websiteButton = document.createElement('button');
   websiteButton.className = 'results-website-button';
-  websiteButton.innerText = 'Visit Website';
-  // TODO: Add website link
+  if (website.includes('maps.google.com')) {
+    websiteButton.innerText = 'Visit Location on Google Maps';
+    linkWebsite(website, websiteButton);
+  }
+  else if (website === '') {
+    websiteButton.innerText = 'Website unavailable';
```
shaquille-hall, line 116: *nit: *Website Unavailable

src/main/webapp/style.css
```diff
@@ -195,6 +195,10 @@ Main Content Container Styling
   color: var(--white);
 }
 
+.unavailable-website {
+  background-color: var(--dark-gray);
```
shaquille-hall, line 199: You may also need to disable clicks here, also you might be getting that for free from the browser because there is no url.

src/main/webapp/script.js
LindseyWeiskopf: fixed!

src/main/webapp/script.js
LindseyWeiskopf: Fixed

src/main/webapp/style.css
LindseyWeiskopf: Disabled clicks!

PR #66, opened 2020-07-23
Made the algorithm to delete big businesses from results less complex

AbakirH:
Summary: I made the method that connects to the database less complex and made it so to delete a business that will be returned from the places API. Also, the method is now of O(n) complexity instead of O(n^2).

I deleted getBigBusinessFromDatabase, because to connect to the database I only need one line of code. Also, instead of a having a nested while loop to compare if the businesses from the results are the same as the big businesses in the database. I made it so each big business now in the database has a key, and I only need to use datastore.get method to access a big business based on its key with one loop and to delete it from the results if found.

### DIFF
--- a/src/test/java/com/google/sps/BigBusinessesTest.java
+++ b/src/test/java/com/google/sps/BigBusinessesTest.java
@@ -41,6 +41,8 @@ import org.junit.Test;
 import static org.junit.Assert.*;
 import static org.hamcrest.CoreMatchers.*;
 import com.google.maps.model.Photo;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
 
 @RunWith(JUnit4.class)
 public final class BigBusinessesTest {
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/sps/BigBusinessesTest.java
```diff
@@ -134,16 +136,19 @@ private void setUpSampleDatabase() {
...
     String photos = "Photos";
     datastore = DatastoreServiceFactory.getDatastoreService();
-    Entity businessEntity = new Entity("BigBusinesses");
+    Entity businessEntity;
+    Key key;
     for(Listing business: sampleDatabaseOfBigBusinesses) {
-      businessEntity = new Entity("BigBusinesses");
+      // key = KeyFactory.createKey("key",business.getName());
+      businessEntity = new Entity("BigBusinesses",business.getName());
       businessEntity.setProperty(title, business.getName());
```
AbakirH, line 143: I think I should delete this line as I am now creating each business with a key associated with it's name. I want to get your opinion on this.

src/test/java/com/google/sps/BigBusinessesTest.java
```diff
@@ -134,9 +136,10 @@ private void setUpSampleDatabase() {
     String rating = "Rating";
     String photos = "Photos";
     datastore = DatastoreServiceFactory.getDatastoreService();
-    Entity businessEntity = new Entity("BigBusinesses");
+    Entity businessEntity;
+    Key key;
     for(Listing business: sampleDatabaseOfBigBusinesses) {
-      businessEntity = new Entity("BigBusinesses");
+      businessEntity = new Entity("BigBusinesses",business.getName());
```
shaquille-hall, line 142: nit: add a space after this comma.

PR #67, opened 2020-07-23
Implement isValidInput() function

LindseyWeiskopf:
Summary: Create function that checks to see if the user entered zipCode is valid.

In this PR, I implemented a function that checks to see if the user entered zip code is valid to prevent the user from being able to input malicious content or alter our javascript/html. 

The function iterates through each character of the input and checks to see if the character's unicode corresponds to a numerical digit, a lowercase or uppercase letter, or a comma, or a space. 

If the input does not follow these parameters, the user gets an alert message and must enter another location.

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -50,7 +50,7 @@ to represent the company, product, or service to which they refer. -->
       <h1 class="logo">smallcity</h1>
       <button class="geolocate-button" onclick="getGeolocation()"><i class="fas fa-location-arrow"></i> locate me</button>
       <div class="zipcode-container">
-        your zipcode:
+        city, state, or zip:
         <input type="text" id="zipCode" name="zipCode" value="">
         <button type="submit" onclick="getZipCode()" class="zipCode-submit-btn"><i class="fas fa-search"></i></button>
 
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -39,7 +39,31 @@ function displayError() {
 
 function getZipCode() {
   let zip = document.getElementById('zipCode').value;
-  fetchList('/data?zipCode=' + zip);
+  if (isValidInput(zip)) {
+    fetchList('/data?zipCode=' + zip);
+  }
+  else {
+    window.alert('Invalid input. Try again.'); 
+  }
+}
+
+function isValidInput(zip) {
+  let len = zip.length;
+  if (len === 0) {  
+    return false; 
+  }
+  let i;
+  for (i=0; i<len; i++) {
+    let charCode = zip.charCodeAt(i);
+    if (!(charCode >= 48 && charCode <= 57) &&      // digits
+          !(charCode >= 65 && charCode <= 90) &&    // uppercase letters
+          !(charCode >= 97 && charCode <= 122) &&   // lowercase letters
+          !(charCode === 44) &&                     // comma
+          !(charCode === 32)) {                     // space
+      return false;
+    }
+  }
+  return true;
 }
 
 //Array of the (currently 6 for this demo build) 15 listings gathered from the fetch request
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -52,8 +52,7 @@ function isValidInput(zip) {
   if (len === 0) {  
     return false; 
   }
-  let i;
-  for (i=0; i<len; i++) {
+  for (let i=0; i<len; i++) {
     let charCode = zip.charCodeAt(i);
     if (!(charCode >= 48 && charCode <= 57) &&      // digits
           !(charCode >= 65 && charCode <= 90) &&    // uppercase letters
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -52,6 +52,7 @@ function isValidInput(zip) {
   if (len === 0) {  
     return false; 
   }
+  
   for (let i=0; i<len; i++) {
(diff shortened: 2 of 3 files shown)


src/main/webapp/script.js
```diff
@@ -39,7 +39,31 @@ function displayError() {
...
+  }
+}
+
+function isValidInput(zip) {
+  let len = zip.length;
+  if (len === 0) {  
+    return false; 
+  }
+  let i;
+  for (i=0; i<len; i++) {
```
shaquille-hall, line 56: Can this be combined to for(let i=0; i<len; i++) ?

LindseyWeiskopf: Yes - just changed!

LindseyWeiskopf:
> Looks good to me, do you want me to change the wording on that text box to something else, considering it can take more than just zipcode?

Ah smart - just changed to "city, state, or zip:"

src/main/webapp/index.html
```diff
@@ -50,7 +50,7 @@
       <h1 class="logo">smallcity</h1>
       <button class="geolocate-button" onclick="getGeolocation()"><i class="fas fa-location-arrow"></i> locate me</button>
       <div class="zipcode-container">
-        your zipcode:
+        city, state, or zip:
```
shaquille-hall, line 53: Is this requesting all of these bits, or some sub combination? What happens if a user only enters the city? What happens if the user only enters the state? What happens if they enter both? etc.

LindseyWeiskopf: It should work if the user enters any combination of those things. I'm not sure how to express that on the front end though...something like "enter your location" or "where are you?" Do you have any suggestions?

AbakirH: You can add filler text inside the zip code search box, to have "Zip code | City | State" and that should be clear for the user.

shaquille-hall: What about cities with the same name, but in different states? What will this return? It seems like one will just be chosen by the api (best case scenario). I think what you have now works fine, but these error cases should be appropriately handled in the code.

LindseyWeiskopf: If there is a cities with the same name, there wouldn't be an issue if the user also entered an addition identifier (like state). If not, I believe that the API would just choose one (ex. if I search "Springfield", the results will be Springfield, MO, but if I search "Springfield, IL" the results will be in Illinois). I'm having a bit of trouble thinking about how these cases can be handled if we don't know what the user wants their results to be. Do you have any suggestions?

shaquille-hall: Before we go there, what benefit is there to allowing users to enter a city, state combination. Results can be returned for presumably anywhere (defined by the API) in that city. These can be outside of the defined radius for the application. Is this outside of the core scope for you app?

LindseyWeiskopf: I don't think that it is. If a user is travelling, they likely won't know the zipcode of the place that they're travelling to off of the top of their head. I think the benefit of the city, state combination is allowing the user to get less specific results while being able to reference a specific city. I think this feature would be more beneficial if we were displaying more than 15 results, but I still think that the use case is valuable.

shaquille-hall: I agree. This is a useful thing to support. My concern however, is that the behaviour seems undefined. i.e. it's not clear what the focal point becomes with only a city & state input pair. Perhaps this is documented in the api docs somewhere?

LindseyWeiskopf: The results of the Geocoding API returns a parameter called ["locationType"](https://developers.google.com/maps/documentation/geocoding/overview#Results) which seems to specify the accuracy of the result. I tested the code a few times, and it seems like "Approximate" is returned for both zipCodes and city/state combinations. When an exact address is entered, it returns "Rooftop" which indicates a precise geocode. One of the other return options is "Geometric center" which to me appears to be what is chosen when a broad input is passed because it looks like it's centering around the location.

shaquille-hall: Ok. You should be able to confirm this in the results payload JSON array. Let's document these various return types in your design doc.

LindseyWeiskopf: I added a logger message to track the location type and I will add to the design doc.

src/main/webapp/script.js
```diff
@@ -39,7 +39,30 @@ function displayError() {
...
+    window.alert('Invalid input. Try again.'); 
+  }
+}
+
+function isValidInput(zip) {
+  let len = zip.length;
+  if (len === 0) {  
+    return false; 
+  }
+  for (let i=0; i<len; i++) {
```
shaquille-hall, line 55: nit: Please add a space above this line.

src/main/webapp/script.js
```diff
@@ -39,7 +39,30 @@ function displayError() {
...
+    let charCode = zip.charCodeAt(i);
+    if (!(charCode >= 48 && charCode <= 57) &&      // digits
+          !(charCode >= 65 && charCode <= 90) &&    // uppercase letters
+          !(charCode >= 97 && charCode <= 122) &&   // lowercase letters
+          !(charCode === 44) &&                     // comma
+          !(charCode === 32)) {                     // space
+      return false;
+    }
+  }
+  return true;
```
shaquille-hall, line 65: nit: Please add a space above this line.

src/main/webapp/script.js
LindseyWeiskopf: Fixed!

PR #69, opened 2020-07-24
Initial JS Tests (paginationScript.js)

ethanimooney:
TL;DR: Added the first of the JS frontend tests, focusing on paginationScript.js
Note: I may have done these terribly if I'm totally honest - please let me know what I can improve and change :) 

**,gitignore** - Added ignore for node_modules/ , as I had to install Node.js to utilize Jest

**package.json package-lock.json** - two files needed for Node and Jest

**paginationScript.test.js** - the actual testing file. Spoke with shaquilleh@ and decided to test the working of the algorithm, rather than the DOM elements themselves. This file contains tests of the main functions, using arrays and objects instead of DOM elements.

Another Note: I also realize these may be lacking, but I wanted to make sure that I get review on the initial ones to make sure I am on the right track.


src/main/webapp/js_tests/paginationScript.test.js
```diff
@@ -0,0 +1,90 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const TOTAL_CARDS_TO_DISPLAY = 3;
+const MAX_LIST_VIEW_NUMBER = 15;
+
+let testResultsContent = [];
+let testResultsCardsList = [];
+
+for (let i = 0; i < 6; i++) {
```
shaquille-hall, line 21: Why 6?

src/main/webapp/js_tests/paginationScript.test.js
```diff
@@ -0,0 +1,90 @@
...
+  expect(testResultsContent[0].id).toEqual(0);
+  expect(testResultsContent[1].id).toEqual(1);
+  expect(testResultsContent[2].id).toEqual(2);
+});
+
+
+test('If the moveNext function works as expected', () => {
+
+  // Move next (what is being tested)
+  testDisplayCards(TOTAL_CARDS_TO_DISPLAY);
```
shaquille-hall, line 74: Let's highlight what the list augment should be here, as opposed to using TOTAL_CARDS_TO_DISPLAY.

src/main/webapp/js_tests/paginationScript.test.js
```diff
@@ -0,0 +1,90 @@
...
+
+  expect(testResultsContent[0].id).toEqual(3);
+  expect(testResultsContent[1].id).toEqual(4);
+  expect(testResultsContent[2].id).toEqual(5);
+});
+
+test(' If the navigatePrevious works as expected', () => {
+
+  // Move Backwards (what is being tested)
+  testDisplayCards((-TOTAL_CARDS_TO_DISPLAY));
```
shaquille-hall, line 84: Same comment as above.

src/main/webapp/js_tests/paginationScript.test.js
```diff
@@ -0,0 +1,90 @@
...
+for (let i = 0; i < 6; i++) {
+  let testCardObject = {
+    id: i,
+    testPhotoReference: ("Card " + i +"\'s Photo Reference")
+  };
+
+  testResultsCardsList.push(testCardObject);
+}
+
+function testDisplayCards(listAugment){
```
shaquille-hall, line 30: Rather than rewriting this logic here, can you reference the displayCards method in pagination.js?

ethanimooney: I would, but the changing from DOM elements to an array of objects wouldn't lend itself to this, so I had to copy and update it for arrays (simple things like appending to an array, etc, no big changes)

src/main/webapp/js_tests/paginationScript.test.js
ethanimooney: I'm just filling the testResultsCardList with 6 cards, enough to test what I need it to.

src/main/webapp/js_tests/paginationScript.test.js
ethanimooney: Gotcha

ethanimooney:
These new tests should work but they will not run on my computer as Jest seems to have cached the old tests, and is trying to run those instead. Need to sync with shaquilleh@ tomorrow to figure this out.

src/main/webapp/paginationScript.test.js
```diff
@@ -0,0 +1,36 @@
...
+  expect(calculateListAugment(TOTAL_CARDS_TO_DISPLAY)).toEqual(0);
+
+  //Some random acceptable location in list, next
+  currentFirstCardIndex = 3;
+  expect(calculateListAugment(TOTAL_CARDS_TO_DISPLAY)).toEqual(3);
+
+  //Some random acceptable location in list, previous
+  currentFirstCardIndex = 3;
+  expect(calculateListAugment(-TOTAL_CARDS_TO_DISPLAY)).toEqual(0);
+});
```
shaquille-hall, line 36: Each of these expects should be a different test. i.e. testMoveNext(), testMovePrevious, testUserAtEndofList etc.

ethanimooney: Done

src/main/webapp/paginationScript.test.js
```diff
@@ -0,0 +1,39 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const calculateListAugment = require('./paginationScript');
+
+const TOTAL_CARDS_TO_DISPLAY = 3;
+const MAX_LIST_VIEW_NUMER = 15;
+
+test('It calculateListAugment performs as expected when at beginning of list', () => {
```
shaquille-hall, line 20: Should this say **"If** calculateListAugment ..."?

src/main/webapp/paginationScript.test.js
```diff
@@ -0,0 +1,39 @@
...
+  currentFirstCardIndex = 0;
```
shaquille-hall, line 21: Do you need this variable? I don't see it being used anywhere. Perhaps you intended for calculateListAugment to receive it as a parameter?

ethanimooney: It's being used in the second test, line 26

ethanimooney: ...Mispelled oops

src/main/webapp/paginationScript.test.js
ethanimooney: Yes, yes it should ... it really is a Friday isn't it

ethanimooney:
As mentioned in the issue for this, the tests are effectively done. I wasn't able to actually run them due to issues with something caching old tests and not running the new ones, which is why this PR hasn't been approved.

PR #70, opened 2020-07-24
Add "Image Not Found" Images

ethanimooney:
TL;DR: Added "image not found" placeholder for businesses that have no images returned from the Places API. 
Note: this change is not live, so the images can be viewed [here](https://screenshot.googleplex.com/wotpuN765KR.png)

(I created two images, one with a white background (one) and one with a dark gray background (two) - went with two, left one just in case it is needed elsewhere)

**paginationScript.js**: Added to if statement that checks if cardToAppend has an image or not that adds the image_not_found_two.png if there is not one.


src/main/webapp/paginationScript.js
```diff
@@ -61,11 +61,14 @@ function displayCards(listAugment) {
       //Card being appended to the resultsContent div
       let cardToAppend = resultsCardsArray[i];
 
+      //The actual image element to which the image src will be applied
+      let imageElement = cardToAppend.card.children[0].children[0];
```
shaquille-hall, line 65: Looks like this is following the append order on line 134 in script.js. 
Can you incorporate the classname  'results-image' here?

src/main/webapp/paginationScript.js
```diff
@@ -61,11 +61,14 @@ function displayCards(listAugment) {
...
 
+      //The actual image element to which the image src will be applied
+      let imageElement = cardToAppend.card.children[0].children[0]; 
       if (cardToAppend.photoReference != 'none') {
-        //The actual image element to which the image src will be applied
-        let imageElement = cardToAppend.card.children[0].children[0];
         loadImage(imageElement, cardToAppend.photoReference);
       }
+      else {
+        imageElement.src = "/images/image_not_found_two.png";
```
LindseyWeiskopf, line 70: nit: single quotes!

ethanimooney: Oops

src/main/webapp/paginationScript.js
ethanimooney: Done

ethanimooney:
I actually ended up moving where the src for where the image not found image is applied. I set it to be applied by default when the card is created. Now, in the pagination script, only if the image has a photoReference, is the source changed.

src/main/webapp/paginationScript.js
shaquille-hall: Was this updated?

ethanimooney: It was, I changed imageElement to resultsImageElement - is this generally what you meant?

shaquille-hall: Not quite. I meant instead of doing "cardToAppend.card.children[0].children[0]" can you find the image element given its class name? That way this is dependent on what order the elements are appended in in script.js.

shaquille-hall:
> I actually ended up moving where the src for where the image not found image is applied. I set it to be applied by default when the card is created. Now, in the pagination script, only if the image has a photoReference, is the source changed.

Good call!

src/main/webapp/paginationScript.js
```diff
@@ -61,21 +61,33 @@ function displayCards(listAugment) {
...
 
+function locateImageElement(card) {
+  let cardChildren = card.childNodes;
+  let child = 0;
+
+  while (cardChildren[child].className != 'results-image') {
+    child++;
+  }
+
+  return cardChildren[child].children[0];
```
shaquille-hall, line 84: Can you grab this by the "img" tag instead? 

This imageElement is a div (that contain many elements). If another element gets added before the 'img' tag, this will likely break. getElementsByTagName seems like it should help here.

ethanimooney: Done!

ethanimooney: I knew there was a better way than that but forgot that existed

PR #76, opened 2020-07-27
Product search implementation

LindseyWeiskopf:
Summary: Implement product search feature using the text Search request and adjusting front end to allow full stack flow of info (with minimal styling at the moment)

I will explain the changes for each file below:

**index.html:** Added entry container which is triggered upon first loading the web app or when the user tried to change their set location. This box contains the option to enter a location or be geolocated (to be stylized with clearer user instructions). Added product search container for the user to enter their product search.

**style.css:** Preliminary styling for new product search features. The attributes used to alter the display of the loader circle were changed to generic so that they can be used for both the loader circle and the entry container.

**script.js:** The most substantial change to this file involves the handling of the query string. Upon first entering the webapp and entering a location, the product query string attribute will be set to 'NULL' which triggers the generic nearbySearchRequest on the back end. If the user enters a product, the product attribute will be set to their entry. For each product and location change, the query string will depend on the values saved as the location and/or product entry. In addition, this file handles the display and hide of the entry container.

**DataServlet.java:** The product is grabbed as a parameter and passed to create the User instance

**UserService.java:** The product is saved as an instance variable to the UserService as it is an attribute of the user's information

**SmallCityService.java:** In the findAllBusinesses() function, the program checks to see whether the product value is "NULL". If so, the NearbySearchRequest function is called. If the product value is a real product, the TextSearchRequest function is called.

**BusinessesService.java:** Added the getBusinessesFromTextSearch() function (very similar to nearbySearch code) and change the Add listing function to add the correct address depending on the function called.

I think that sums it all up -- Thanks!


src/main/webapp/index.html
```diff
@@ -43,17 +43,30 @@
...
+        your zipcode:
+        <input type="text" id="entryZipCode" name="entryZipCode" value="">
+        <button type="submit" onclick="getZipCode()" class="zipCode-submit-btn"><i class="fas fa-search"></i></button>
+      </div>
+    </div>
+
+    
     <div id="loader-circle-container">
       <div id="loader-circle"></div>
     </div>
```
ethanimooney, line 59: I believe here we pulled out loader-circle and put it before loader-circle-container to make it so the loader wasn't transparent. If you want to make this change you might need to rename loader-circle-container to something different.

This move can also be done for the popup I think

src/main/webapp/script.js
```diff
@@ -14,9 +14,12 @@
 
 const alertMessage = 'Sorry! We cannot geolocate you. Please enter a zipcode';
 let map;
+let locationQuery = '';
+let product = '';
```
ethanimooney, line 18: I might change this to productQuery, to keep it consistent, unless there is another reason for this difference.

ethanimooney: Oh wait I now see the reason, disregard haha

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -44,10 +44,12 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     double lat = convertToDouble(latString);
     double lng = convertToDouble(lngString);
     MapLocation userLocation = new MapLocation(lat, lng);
-    smallCityService.createUserServiceWithGeolocation(userLocation);
+    String product = request.getParameter("product");
+    smallCityService.createUserServiceWithGeolocation(userLocation, product);
     } catch(NullPointerException e) {
```
ethanimooney, line 49: nit: Hate to be that guy but need a space after catch

src/main/webapp/index.html
LindseyWeiskopf: You're totally right - just fixed it

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: Fixed

LindseyWeiskopf:
> Looks pretty good, wondering if you could explain the reasoning for saving the product as an instance variable in the user though? (I'm probably just missing something here)

I think this was part of our original play (to make the product value a part of the user object) because the search technically belongs to the user. It also makes it easier to reference the product as input on the backend when making the text search request. Does that make sense?

ethanimooney:
> > Looks pretty good, wondering if you could explain the reasoning for saving the product as an instance variable in the user though? (I'm probably just missing something here)
> 
> I think this was part of our original play (to make the product value a part of the user object) because the search technically belongs to the user. It also makes it easier to reference the product as input on the backend when making the text search request. Does that make sense?

Ahh yes yes - remember that now. It sure is a Monday, isn't it...

src/main/java/com/google/sps/data/SmallCityService.java
```diff
@@ -38,23 +38,28 @@ public SmallCityService() { }
   * Create UserService instance from zipCode and get businesses list
   * @param zipCode inputted zipcode of user
   **/
-  public void createUserServiceWithZip(String zipCode) {
+  public void createUserServiceWithZip(String zipCode, String product) {
     this.userService = new UserService();
-    userService.createUserWithZipCode(zipCode);
+    userService.createUserWithZipCode(zipCode, product);
```
shaquille-hall, line 43: I believe you have a SearchObject class that would be much better suited for this - containing the product search string that was use. 
ref: https://github.com/googleinterns/SmallCity/blob/master/src/main/java/com/google/sps/data/SearchObject.java

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -68,7 +69,36 @@ public BusinessesService(List<Listing> allBusinesses) {
...
-  public List<Listing> getBusinessesFromPlacesApi(MapLocation mapLocation) {
+  public List<Listing> 
+        getBusinessesFromTextSearch(MapLocation mapLocation, String product) {
+    LatLng latLng = 
+          new LatLng(mapLocation.lat, mapLocation.lng);
+    final GeoApiContext context = new GeoApiContext.Builder()
+            .apiKey(KEY)
+            .build();
+    TextSearchRequest request = PlacesApi.textSearchQuery(context, product);
+    try {
```
shaquille-hall, line 80: nit: Please add a space above this line to improve readability.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -68,7 +69,36 @@ public BusinessesService(List<Listing> allBusinesses) {
...
+      PlacesSearchResponse response = request.location(latLng)
+              .radius(10000)
```
shaquille-hall, line 82: Please make 10000 a constant

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -68,7 +69,36 @@ public BusinessesService(List<Listing> allBusinesses) {
...
+              .await();
+      for (int i=0; i<ALLOWED_SEARCH_REQUESTS; i++) {
```
shaquille-hall, line 84: nit: Please add a space above this line to improve readability.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -44,10 +44,12 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     double lng = convertToDouble(lngString);
     MapLocation userLocation = new MapLocation(lat, lng);
-    smallCityService.createUserServiceWithGeolocation(userLocation);
-    } catch(NullPointerException e) {
+    String product = request.getParameter("product");
+    smallCityService.createUserServiceWithGeolocation(userLocation, product);
+    } catch (NullPointerException e) {
       String zip = request.getParameter("zipCode");
-      smallCityService.createUserServiceWithZip(zip);
+      String product = request.getParameter("product");
```
shaquille-hall, line 51: Odd. This is a repeat of line 47.

src/main/webapp/script.js
```diff
@@ -29,7 +32,14 @@ function getGeolocation() {
 function displayLocation(position) {
   let lat = position.coords.latitude;
   let lng = position.coords.longitude;
-  fetchList('/data?lat=' + lat + "&lng=" + lng);
+  locationQuery = '/data?lat=' + lat + '&lng=' + lng;
+  product = document.getElementById('product').value;
+  if (product === '') {
+    fetchList(locationQuery + '&product=NULL');
+  }
```
shaquille-hall, line 39: Would it be simpler to leave this empty. and use .equals("") in SmallCityService?

src/main/webapp/script.js
```diff
@@ -38,8 +48,41 @@ function displayError() {
...
+  hideEntryContainer();
+  initiateLoaderCircle();
+  zip = document.getElementById('entryZipCode').value;
+  document.getElementById('zipCode').value = zip;
+  locationQuery = '/data?zipCode=' + zip;
+  product = document.getElementById('product').value;
+  if (product === '') {
+    fetchList(locationQuery + '&product=NULL');
+  }
+  else {
```
shaquille-hall, line 60: Same comment as above. Also, given this is the same logic as above, perhaps move it to a different method, and call that where relevant.

src/main/webapp/script.js
```diff
@@ -38,8 +48,41 @@ function displayError() {
...
+
+function getProduct() {
+  initiateLoaderCircle();
+  if (document.getElementById('product').value === '') {
+    fetchList(locationQuery + '&product=NULL');
+  }
+  else {
+    product = document.getElementById('product').value;
+    fetchList(locationQuery + '&product=' + product);
+  }
```
shaquille-hall, line 73: Same comment as above.

src/main/java/com/google/sps/data/SmallCityService.java
LindseyWeiskopf: I implemented the SearchObject class and removed the instance from the user service. Now, the searchObject only exists in the DataServlet and smallCityService.

src/main/java/com/google/sps/data/BusinessesService.java
LindseyWeiskopf: Done

src/main/java/com/google/sps/data/BusinessesService.java
LindseyWeiskopf: Fixed.

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: It's a repeat because only one block of code will be executed. If there is no geolocation, the try block will fail before the product is grabbed.

src/main/webapp/script.js
LindseyWeiskopf: yes it is. I just changed it.

src/main/webapp/script.js
LindseyWeiskopf: I refactored this bit of code into a separate function (fetchByQueryString()).

src/main/java/com/google/sps/servlets/DataServlet.java
shaquille-hall: Right, but this is a code smell. This is retrying logic in the catch block that was already attempted in the try block - which defeats the purpose of the try/catch. Looks like the NPE is thrown from request.getParameter(). How about centering the try/catch only around those calls, something like the following:
```
public void doGet() {
    String latString;
    String product;
    ....
    
    try {
        latString = request.getParameter("lat");
        product = request.getParameter("product");
    } catch (NPE e) {
        Logger.warning(e.getMessage() + "more explanation");
    }

    SearchObject searchObject = new searchObject(product);
    smallCirtService.createUser(...);
    .....
}
```
An if statement might also be helpful.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -44,10 +45,14 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     double lat = convertToDouble(latString);
     double lng = convertToDouble(lngString);
     MapLocation userLocation = new MapLocation(lat, lng);
-    smallCityService.createUserServiceWithGeolocation(userLocation);
-    } catch(NullPointerException e) {
+    String product = request.getParameter("product");
+    SearchObject searchObject = new SearchObject(product);
+    smallCityService.createUserServiceWithGeolocation(userLocation, searchObject);
+    } catch (NullPointerException e) {
```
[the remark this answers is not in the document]
LindseyWeiskopf: Done!

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: I restructured to a try/catch/finally to make sure that the right code is still being executed after the catch while repeating less code. Let me know if this solution still needs work!

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -36,23 +38,34 @@
...
-    } catch(NullPointerException e) {
-      String zip = request.getParameter("zipCode");
-      smallCityService.createUserServiceWithZip(zip);
+      latString = request.getParameter("lat");
+      lngString = request.getParameter("lng");
+      product = request.getParameter("product");
+    } catch (NullPointerException e) {
       LOGGER.warning(e.getMessage() 
            + "Unable to geolocate user, zipCode entered instead.");
+    } finally {
```
shaquille-hall, line 52: I don't think you gain anything with a finally block here. Here is a great resource on it: https://www.geeksforgeeks.org/flow-control-in-try-catch-finally-in-java/

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -36,23 +38,34 @@
...
+      SearchObject searchObject = new SearchObject(product);
+      if (latString != null && lngString != null) {
```
shaquille-hall, line 54: This should probably check for null and isEmpty. How does MapLocation work with an empty lat or lng value?

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: Yes you're right - I removed it.

src/main/java/com/google/sps/servlets/DataServlet.java
LindseyWeiskopf: I added the extra checks - MapLocation does not do much in itself but I would imagine that the Places API calls would fail with empty coordinates.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -36,20 +38,32 @@
...
       LOGGER.warning(e.getMessage() 
            + "Unable to geolocate user, zipCode entered instead.");
     }
-  
+      
+    SearchObject searchObject = new SearchObject(product);
+    if (latString != null && !latString.isEmpty() 
+          && lngString != null && !lngString.isEmpty()) {
+      double lat = convertToDouble(latString);
+      double lng = convertToDouble(lngString);
```
shaquille-hall, line 58: The guava library has a nifty method that simplifies this, but this also works. 
If you're curious: https://guava.dev/releases/19.0/api/docs/com/google/common/base/Strings.html#isNullOrEmpty(java.lang.String)

LindseyWeiskopf: Interesting! Decided to use this instead

PR #77, opened 2020-07-28
Cache small businesses using Google's datastore API

AbakirH:
Summary: Able to shorten the load time of our project by storing the small business within a database.
Also, fixed readability of the code and how it all flowed together.

**BusinessesService.java:**  

Reformatted how the code removes a big business from the list returned from the places API:
- removeBigBusinessesFromResults() - This function now calls determineIfTheCurrentBusinessIsBig() and if it returns true the business is removed and if it returns false, I add one to the numberOfSmallBusinesses variable. I do this so we only get the first 15 small businesses and display on the page.
- determineIfTheCurrentBusinessIsBig() - 
  1.  Checks if the business is in the SmallBusiness database, if so we do not need to make any additional checks. - checkIfBusinessInDatabase() 
  2. Checks if business is in the BigBusiness database, if it is the program returns true to remove that business from the list and it does not make any additional checks. - checkIfBusinessInDatabase() 
  3. Check if the business is big using the checkNumberOfSimilarBusinesses() and checkBusinessThroughLinkedin(). If it is determined to be a big business, the program adds the business to the big business database using the addBusinessToDatabase() method and also return true, to remove the it from the appropriate database.
  4. Otherwise the business is small and use the addBusinessToDatabase() to add the business to the SmallBusiness database and returns false.

- checkNumberOfSimilarBusinesses() - returns true or false if the business is determined to be a big business calling the checkNumberOfSimilarBusinessesInTheArea()

- checkNumberOfSimilarBusinessesInTheArea - Loops through the results from the similarBusinessesInTheArea variable and if more then 5 matching businesses are found, returns true for that business to be added to the appropriate database and removed from the list.

- addBusinessToDatabase() - takes an additional String variable parameter, for the program to determine which database to add the business information to.

**SmallCityService.java:**  
 
- findBigBusinessInList() - removed this function as it was determined it was not needed as all the checks to determine if a business is big or small should happen from the filterBySmallBusinesses() method.

### DIFF
--- a/src/main/java/com/google/sps/data/BusinessesService.java
+++ b/src/main/java/com/google/sps/data/BusinessesService.java
@@ -61,7 +61,6 @@ public class BusinessesService {
   private final String SMALL_BUSINESSES_DATABASE = "SmallBusinesses";
   private LatLng latLng;
   private List<Listing> allBusinesses;
-  private int numberOfSmallBusinesses;
 
   /** Create a new Businesses instance
   * @param allBusinesses businesses from SmallCityService
@@ -75,7 +74,8 @@ public class BusinessesService {
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     Iterator<Listing> businesses =  allBusinesses.iterator();
     String businessName;
-    while (businesses.hasNext()) {
+    int numberOfSmallBusinesses = 0;
+    while (businesses.hasNext() && numberOfSmallBusinesses < SMALL_BUSINESSES_DISPLAYED) {
       Listing currentBusiness = businesses.next();
       try {
         businessName = 
@@ -87,7 +87,11 @@ public class BusinessesService {
           businesses.remove();
         }
       } catch(EntityNotFoundException e){
-        LOGGER.warning(e.getMessage());
+         if(determineIfTheBusinessesAreBig(currentBusiness)) {
+           businesses.remove();
+         }else{
+            numberOfSmallBusinesses++;
+         }
       }
     }
     return allBusinesses;
@@ -151,38 +155,31 @@ public class BusinessesService {
           placeLocation, rating, photos, types, url));
   }
 
-  public void determineIfTheBusinessesAreBig() {
-    numberOfSmallBusinesses = 0;
-    Iterator<Listing> businesses =  allBusinesses.iterator();
-    boolean isBigBusiness;
-    while(businesses.hasNext()
-          && numberOfSmallBusinesses < SMALL_BUSINESSES_DISPLAYED) {
-      Listing currentBusiness = businesses.next();
-
-      isBigBusiness = checkIfSmallBusinessInDatabase(currentBusiness);
-      
-      if(isBigBusiness){
-        businesses.remove();
-      }
+  public boolean determineIfTheBusinessesAreBig(Listing currentBusiness) {
+    boolean isBigBusiness = false;
+    if(checkIfSmallBusinessInDatabase(currentBusiness)) {
+      isBigBusiness = false;
+    }else {
+      isBigBusiness = checkNumberOfSimilarBusinesses(currentBusiness);
     }
+    return isBigBusiness;
   }
 
   private boolean checkIfSmallBusinessInDatabase(Listing currentBusiness){
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();    
-    boolean isBigBusiness = false;
+    boolean foundSmallBusiness = false;
       try {
           String smallBusinessName = 
               (String) datastore.get(KeyFactory.createKey(
                                       SMALL_BUSINESSES_DATABASE, 
                                       currentBusiness.getName()))
                                 .getProperty("Business");
-          numberOfSmallBusinesses++;
-          isBigBusiness = false;
-          
+          foundSmallBusiness = true;
       } catch(EntityNotFoundException e) {
-          isBigBusiness = checkNumberOfSimilarBusinesses(currentBusiness);
+          LOGGER.warning(e.getMessage());
+          foundSmallBusiness = false;
         }
-    return isBigBusiness;
+    return foundSmallBusiness;
   }
 
   private boolean checkNumberOfSimilarBusinesses(Listing currentBusiness){
@@ -206,8 +203,6 @@ public class BusinessesService {
         isBigBusiness = checkBusinessThroughLinkedin(currentBusiness, similarBusinessesInTheArea);
       } else if(similarBusinessesInTheArea.length == 1) {
         addBusinessToDatabase(currentBusiness, SMALL_BUSINESSES_DATABASE);
-        numberOfSmallBusinesses++;
-        isBigBusiness = false;
       }
     } catch(GeneralSecurityException | IOException | InterruptedException | ApiException e ) {
         LOGGER.warning(e.getMessage());
@@ -280,8 +275,6 @@ public class BusinessesService {
         isBigBusiness = true;
      } else {
         addBusinessToDatabase(currentBusiness, SMALL_BUSINESSES_DATABASE);
-        numberOfSmallBusinesses++;
-        isBigBusiness = false;
      }
     return isBigBusiness;
    }
--- a/src/main/java/com/google/sps/data/BusinessesService.java
+++ b/src/main/java/com/google/sps/data/BusinessesService.java
@@ -1,11 +1,5 @@
 package com.google.sps.data;
 
-import java.util.List;
-import java.util.LinkedList;
-import java.util.Arrays;
-import java.io.Serializable;
-import java.net.URL;
-import java.util.logging.Logger;
 import com.google.maps.GeoApiContext;
 import com.google.maps.PlacesApi;
 import com.google.maps.NearbySearchRequest;
@@ -19,8 +13,6 @@ import com.google.maps.model.PlacesSearchResponse;
 import com.google.maps.model.PlacesSearchResult;
 import com.google.maps.model.PlaceDetails;
 import com.google.maps.PlaceDetailsRequest;
-import java.util.logging.Logger;
-import java.util.Iterator;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -57,10 +57,12 @@
   private final String START_SUBSTRING = "| ";
   private final String END_SUBSTRING = "followers";
   private final int ALLOWED_NUMBER_OF_MATCHING_BUSINESSES = 5;
+  private final String BIG_BUSINESSES_DATABASE = "BigBusinesses";
+  private final String SMALL_BUSINESSES_DATABASE = "SmallBusinesses";
   private LatLng latLng;
   private List<Listing> allBusinesses;
   private int numberOfSmallBusinesses;
-  
+  private boolean isBigBusiness;
```
shaquille-hall, line 65: As written, this says your BusinessesService is a big business (or is not, depending on the value of this boolean). This is an anti-pattern in object oriented programming. A better design here would be to use a function. A function that take a Listing object, executes the three checks that are being used to determine whether or not a business is big, and returns true/false based on that. You can then use that on line 163 in the same way that you're using the boolean.

AbakirH: I changed the flow of the program so that the functions return if the business is big or not. The only problem I am facing is that since you can not return something within an if statement without returning something at the end of the function. I had to make multiple boolean variables within each function to either return true or false at the end of each appropriate function.

shaquille-hall: I'm not seeing any updates in the code. Can you try re-pushing your commit?

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -272,6 +283,7 @@ private void checkNumberOfSimilarBusinessesInTheArea(Listing currentBusiness,
         numberOfSmallBusinesses++;
         isBigBusiness = false;
      }
+    return isBigBusiness;
```
AbakirH, line 286: This is an example of the change I made

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: The code is up to date, look at the commit under this comment. (changed function to return true or false if the business is big)

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -150,41 +151,74 @@ private void addListingToBusinesses(PlacesSearchResult place, String url) {
           placeLocation, rating, photos, types, url));
   }
 
-  public void checkIfBusinessesAreBig() {
-    GeoApiContext context = 
-      new GeoApiContext.Builder().apiKey(KEY).build();
+  public void determineIfTheBusinessesAreBig() {
```
shaquille-hall, line 154: I think you have two methods that imply they're doing the same work, but they're acting independently: removeBigBusinessesFromResults(), and determineIfBusinessesAreBig(). It seems, to me that determineIfBusinessesAreBig() should be a call from removeBigBusinessesFromResults(). i.e
```
removeBigBusinessesFromResults() {
  DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
  Iterator<Listing> businesses =  allBusinesses.iterator();
  String businessName;
  
  while (businesses.hasNext()) {
   Listing currentBusiness = businesses.next();
   try {
        businessName = 
            (String) datastore.get(KeyFactory.createKey(
                                    BIG_BUSINESSES_DATABASE, 
                                    currentBusiness.getName()))
                              .getProperty("Business");

    if (businessName.equals(currentBusiness.getName()) || determineIfTheBusinessesAreBig(currentBusiness) {
        businesses.remove(currentBusiness)
    }
....
}
```

With that determineIfTheBusinessesAreBig() should be refactored to only consider one business at a time. This will also allow you to remove the while loop that's in determineIfTheBusinessAreBig() because the notion of currentBusiness will then be passed in as a parameter.

shaquille-hall: Following that, determineIfBusinessesAreBig() can become a simple wrapper method for your actual checks: checkIfSmallBusinessInDatabase(), checkNumberOfSimilarBusinesses(), checkBusinessThroughLinkedin():
```
private boolean determineIfBussinessIsBig(Listing currentBusiness) {
    return (checkIfSmallBusinessInDatabase(currentBusiness) &&
        checkNumberOfSimilarBusinesses(currentBusiness) && 
        checkBusinessThroughLinkedin(currentBusiness))
}
```

src/main/java/com/google/sps/data/SmallCityService.java
```diff
@@ -69,8 +69,7 @@ public void filterBySmallBusinesses() {
   }
   
   public void findBigBusinessInList () {
-    businessesService.checkIfBusinessesAreBig();
-    filterBySmallBusinesses();     
+    businessesService.determineIfTheBusinessesAreBig();    
   }
```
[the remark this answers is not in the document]
AbakirH: I can see where you are coming from and I agree with your statement and that is what I did, but in a slightly different way

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: I can not write the if statement like you did because, since I am using a try catch block. If the big business is not in the database it follows the code inside the catch statement. So I decided to make a if statement that uses determineIfTheBusinessesAreBig() to delete the business in the same way. 
`catch(EntityNotFoundException e){
    if(determineIfTheBusinessesAreBig(currentBusiness)) {
      businesses.remove();
     }
  }`

However, the problem with what you wrote within determineIfTheBusinessesAreBig() is that I do not want to do the other checks if the business is found in the SmallBusiness database. In order to do this, I used a if else statement as follows: 
`  public boolean determineIfTheBusinessesAreBig(Listing currentBusiness) {
     boolean isBigBusiness = false;
     if(checkIfSmallBusinessInDatabase(currentBusiness)) {
       isBigBusiness = false;
     }else {
       isBigBusiness = checkNumberOfSimilarBusinesses(currentBusiness);
     }
     return isBigBusiness;
  }`
Also, since the checks are dependent of each other, I only should call one as they call each other if one check is not sufficient enough to determine if a business is big.

src/main/java/com/google/sps/data/SmallCityService.java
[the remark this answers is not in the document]
AbakirH: I discussed this more with Shaquille and highlighted what I changed in the program in the description of the pull request - https://github.com/googleinterns/SmallCity/pull/77#issue-457938649

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -150,41 +151,74 @@ private void addListingToBusinesses(PlacesSearchResult place, String url) {
...
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();    
+    boolean isBigBusiness = false;
+      try {
+          String smallBusinessName = 
+              (String) datastore.get(KeyFactory.createKey(
+                                      SMALL_BUSINESSES_DATABASE, 
+                                      currentBusiness.getName()))
+                                .getProperty("Business");
           numberOfSmallBusinesses++;
+          isBigBusiness = false;
```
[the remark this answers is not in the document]
AbakirH: I discussed this more with Shaquille and highlighted what I changed in the program in the description of the pull request - https://github.com/googleinterns/SmallCity/pull/77#issue-457938649

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: I discussed this more with Shaquille and highlighted what I changed in the program in the description of the pull request - https://github.com/googleinterns/SmallCity/pull/77#issue-457938649

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -150,109 +128,156 @@ private void addListingToBusinesses(PlacesSearchResult place, String url) {
...
+            (String) datastore.get(KeyFactory.createKey(
+                                    databaseEntry, 
+                                    currentBusiness.getName()))
+                              .getProperty("Business");
+        return true;
+    } catch (EntityNotFoundException e) {
+        // Sometimes the small or big business will be in the database,
+        // but if not additional checks will follow to determine it's size.
+        LOGGER.warning(e.getMessage());
+        return false;
```
shaquille-hall, line 182: For consistency, move this outside of the catch block to the end of the function. It won't change any behaviour.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -150,109 +128,156 @@ private void addListingToBusinesses(PlacesSearchResult place, String url) {
...
+    return checkNumberOfSimilarBusinessesInTheArea(currentBusiness, 
+                                                  similarBusinessesInTheArea);  
+  }
+
+  private boolean checkNumberOfSimilarBusinessesInTheArea(Listing currentBusiness, 
+                          PlacesSearchResult[] similarBusinessesInTheArea) {
+    int numberOfMatchingBusinesses = 0;
+    int i = 0;
+
+    while (i < similarBusinessesInTheArea.length
```
shaquille-hall, line 209: Thanks for moving these methods right after each other!

If the try on line 194/195 fails, I think this will through a null error. A safe resolution would be to initialize similarBusinessesInTheArea to an empty list instead. Then if the try fails, the while loop never executes and you return false. This seems appropriate given there would be no similar businesses to consider.

src/main/java/com/google/sps/data/SmallCityService.java
```diff
@@ -18,10 +11,17 @@
...
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.LinkedList;
+import java.util.Arrays;
+import java.io.Serializable;
+import java.net.URL;
+import java.util.logging.Logger;
+import java.util.Iterator;
```
shaquille-hall, line 24: Most of these are not being used. Please review.

AbakirH: I fixed this, but I was planning to go through all the import statements in another pull request and remove the ones that were not used. I went through all the files and deleted the import statements that were not used.

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Thank you for this suggestion, this makes sense.

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Yea I forgot to change it to an empty list, but it should be good now

src/main/java/com/google/sps/data/UserService.java
```diff
@@ -1,14 +1,8 @@
 package com.google.sps.data;
 
 import com.google.maps.GeocodingApi;
-import com.google.maps.model.AddressComponentType;
-import com.google.maps.model.AddressType;
-import com.google.maps.model.ComponentFilter;
 import com.google.maps.model.GeocodingResult;
 import com.google.maps.GeoApiContext;
-import com.google.maps.model.LatLng;
```
LindseyWeiskopf, line 9: Thank you for deleting all of the extra imports! I'm pretty sure that this import is needed for this call to work (LatLng)

AbakirH: If you are not using the LatLng within the file, you do not need the import and I tested the code out and it was working fine after I deleted all the imports.

LindseyWeiskopf: Hmm interesting you're right. The object is referenced as part of the result but I guess since we're not creating a LatLng object you don't the import. Disregard!

PR #81, opened 2020-07-30
Storing listings and location of the user for their local search

AbakirH:
Summary: Able to store the listings of all the business the user just searched for and location on their machine only. 

listingsLocalStorage variable  was created to store the listings of the latest fetch of all the business returned from the back end.

window.onbeforeunload function - allows for the web application to store the most recent listings of all business and the location they searched with. I am also storing the zip code that the coordinate points correlate to. Also depending if the user searched for a product, that also is saved 

window.onload function - checks if their are listings of businesses that you already searched for before hand and the location you used to get those results. This allows for the user not to have to find the last location they searched up for. The nearest zip code from the latest search is also stored that displays in the your location bar. If a product was searched for, that results shows in the product search bar.

reverseGeocodeGeolocation() - got rid of the "near" string for the your location input box when a user uses the locate me button.


src/main/webapp/script.js
```diff
@@ -249,3 +257,27 @@ function createRating(rating) {
...
+  locationQuery = localStorage.getItem("location");
+  bounds = new google.maps.LatLngBounds();
+  let zipcode = localStorage.getItem("zipcode");
+  let product = localStorage.getItem("product");
+
+  if(listingsLocalStorage != null && locationQuery != null && zipcode != null){
+    hideEntryContainer();
+    document.getElementById('zipCode').value = zipcode;
+    addResultCardsAndMapToTheScreen(listingsLocalStorage);
+    if (product !== "") {
```
shaquille-hall, line 279: Should locationQuery be mentioned in this block somewhere? It's in the if statement, but not used.

AbakirH: Even though it is not being used, I do not want the page to load the most recent list of businesses, without have the locationQuery as well

shaquille-hall: Got it. Sgtm.

src/main/webapp/script.js
```diff
@@ -249,3 +257,27 @@ function createRating(rating) {
...
+}
+
+window.onload = function() {
+  listingsLocalStorage = JSON.parse(localStorage.getItem("listings"));
+  locationQuery = localStorage.getItem("location");
+  bounds = new google.maps.LatLngBounds();
+  let zipcode = localStorage.getItem("zipcode");
+  let product = localStorage.getItem("product");
+
+  if(listingsLocalStorage != null && locationQuery != null && zipcode != null){
```
shaquille-hall, line 275: nit: Please add spaces here. It should be 
```
 if (listingsLocalStorage != null && locationQuery != null && zipcode != null) {
```

AbakirH: fixed

PR #83, opened 2020-07-31
Make closer businesses that have the searched product appear

AbakirH:
Summary: I made the radius smaller for the product search for the Text Search Request. Also, if enough results are not returned, another request is made, but with a bigger radius.

I changed the TEXT_SEARCH_RADIUS variable into the MAX_ALLOWED_TEXT_SEARCH_RADIUS variable which will be used to stop the changing radius if needed.

getBusinessesFromTextSearch() - Inside this method, I made a local radius variable and it is used for the product search using the text search request. However, if not enough results are returned using that radius, I will quadruple that radius and make another request using it. If needed this process will repeat until, either enough results are returned or that, radius is bigger then the MAX_ALLOWED_TEXT_SEARCH_RADIUS.

### DIFF
--- a/src/main/java/com/google/sps/data/BusinessesService.java
+++ b/src/main/java/com/google/sps/data/BusinessesService.java
@@ -90,7 +90,7 @@ public class BusinessesService {
       while (response.results.length < MAX_NUMBER_OF_RESULTS_PER_REQUEST && 
              radius < MAX_ALLOWED_TEXT_SEARCH_RADIUS) {
         // Increase radius if there are less than 20 results 
-        radius = INITIAL_RADIUS * RADIUS_MULTIPLIER;
+        radius *= RADIUS_MULTIPLIER;
         response = request.location(latLng).radius(radius).await();
       }
 
--- a/src/main/java/com/google/sps/data/BusinessesService.java
+++ b/src/main/java/com/google/sps/data/BusinessesService.java
@@ -77,7 +77,7 @@ public class BusinessesService {
             .apiKey(KEY)
             .build();
     TextSearchRequest request = PlacesApi.textSearchQuery(context, product);
-    int radius = 500;
+    int radius = 1500;
 
     try {
       PlacesSearchResponse response = request.location(latLng)
--- a/src/main/java/com/google/sps/data/BusinessesService.java
+++ b/src/main/java/com/google/sps/data/BusinessesService.java
@@ -52,6 +52,8 @@ public class BusinessesService {
   
   private final int ALLOWED_SEARCH_REQUESTS = 3;
   private final int MAX_ALLOWED_TEXT_SEARCH_RADIUS = 50000;
+  private final int RADIUS_MULTIPLIAR = 4;
+  private final int MAX_NUMBER_OF_RESULTS_PER_REQUEST = 20;
   private final int MIN_FOLLOWERS = 50000;
   private final int SMALL_BUSINESSES_DISPLAYED = 15;
   private final String START_SUBSTRING = "| ";
@@ -84,9 +86,9 @@ public class BusinessesService {
               .radius(radius)
               .await();
 
-      while (response.results.length < 20 && 
+      while (response.results.length < MAX_NUMBER_OF_RESULTS_PER_REQUEST && 
              radius < MAX_ALLOWED_TEXT_SEARCH_RADIUS) {
-        radius *= 4;
+        radius *= RADIUS_MULTIPLIAR;
         response = request.location(latLng).radius(radius).await();
       }
 
--- a/src/main/java/com/google/sps/data/BusinessesService.java
+++ b/src/main/java/com/google/sps/data/BusinessesService.java
@@ -52,7 +52,7 @@ public class BusinessesService {
   
   private final int ALLOWED_SEARCH_REQUESTS = 3;
   private final int MAX_ALLOWED_TEXT_SEARCH_RADIUS = 50000;
-  private final int RADIUS_MULTIPLIAR = 4;
+  private final int RADIUS_MULTIPLIER = 4;
   private final int MAX_NUMBER_OF_RESULTS_PER_REQUEST = 20;
   private final int MIN_FOLLOWERS = 50000;
   private final int SMALL_BUSINESSES_DISPLAYED = 15;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -77,12 +77,19 @@ public BusinessesService(List<Listing> allBusinesses) {
...
     try {
       PlacesSearchResponse response = request.location(latLng)
-              .radius(TEXT_SEARCH_RADIUS)
+              .radius(radius)
               .await();
-      
+
+      while (response.results.length < 20 && 
+             radius < MAX_ALLOWED_TEXT_SEARCH_RADIUS) {
+        radius *= 4;
```
shaquille-hall, line 89: Please make 20 and 4 well-defined constants at the top. A comment may also be useful.

AbakirH: fixed

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -51,7 +51,9 @@
         Logger.getLogger(BusinessesService.class.getName());
   
   private final int ALLOWED_SEARCH_REQUESTS = 3;
-  private final int TEXT_SEARCH_RADIUS = 10000;
+  private final int MAX_ALLOWED_TEXT_SEARCH_RADIUS = 50000;
+  private final int RADIUS_MULTIPLIAR = 4;
```
shaquille-hall, line 55: *RADIUS_MULTIPLIER

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -77,12 +79,21 @@ public BusinessesService(List<Listing> allBusinesses) {
             .apiKey(KEY)
             .build();
     TextSearchRequest request = PlacesApi.textSearchQuery(context, product);
-    
+    int radius = 1500;
```
shaquille-hall, line 82: Please make 1500 a constant. Sorry I missed this in my last review. Also, why is this radius different from the already named radius above?

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Thanks for catching this

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: 1500 is not a constant, it is the initial radius for the text search request and if enough results are not returned, I increase it by the RADIUS_MULTIPLIER. However, the MAX_ALLOWED_TEXT_SEARCH_RADIUS variable is the max number that the radius can be for the text search request. I make sure the current radius is not greater then this just in case someone puts in a zip code that naturally has not that many results will allow me not to make to many requests

shaquille-hall: RIght. This is currently an integer literal. Given that it has some meaning to the code I am suggesting it be converted to a defined constant to provide that context. In both cases the value will not change. Here is a reference: https://www.diffen.com/difference/Constant_vs_Literal

In regards to the varying radii, I see. Thank you for clarifying. Understanding this is a starting radius, it can still be converted to a constant :)

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -77,12 +79,21 @@ public BusinessesService(List<Listing> allBusinesses) {
...
+              .radius(radius)
               .await();
-      
+
+      while (response.results.length < MAX_NUMBER_OF_RESULTS_PER_REQUEST && 
+             radius < MAX_ALLOWED_TEXT_SEARCH_RADIUS) {
+        // If the list of businesses are not enough, the radius is increased
+        // in hopes to get more results for that product search      
+        radius *= RADIUS_MULTIPLIER;
+        response = request.location(latLng).radius(radius).await();
```
LindseyWeiskopf, line 94: nit: let's try to make the formatted consistent with the response on line 85 and line 94 (either way)

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -77,12 +79,21 @@ public BusinessesService(List<Listing> allBusinesses) {
...
     try {
       PlacesSearchResponse response = request.location(latLng)
-              .radius(TEXT_SEARCH_RADIUS)
+              .radius(radius)
               .await();
-      
+
+      while (response.results.length < MAX_NUMBER_OF_RESULTS_PER_REQUEST && 
+             radius < MAX_ALLOWED_TEXT_SEARCH_RADIUS) {
+        // If the list of businesses are not enough, the radius is increased
```
LindseyWeiskopf, line 91: nit: the wording of this comment is a bit awkward - I'm not sure the comment is necessary. I think it could be something as simple as "Increase radius if there are less than 20 results"

AbakirH: Thank you for this and I have changed the comment

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Since the line is less then 80 characters it might be better to leave it as it is

src/main/java/com/google/sps/data/BusinessesService.java
AbakirH: Okay I think I was confused what you meant by make it a constant variable. However, I see now that you wanted me to make the number 1500 a constant and use it within the function.

src/main/java/com/google/sps/data/BusinessesService.java
```diff
@@ -77,12 +80,20 @@ public BusinessesService(List<Listing> allBusinesses) {
...
       PlacesSearchResponse response = request.location(latLng)
-              .radius(TEXT_SEARCH_RADIUS)
+              .radius(radius)
               .await();
-      
+
+      while (response.results.length < MAX_NUMBER_OF_RESULTS_PER_REQUEST && 
+             radius < MAX_ALLOWED_TEXT_SEARCH_RADIUS) {
+        // Increase radius if there are less than 20 results 
+        radius = INITIAL_RADIUS * RADIUS_MULTIPLIER;
```
shaquille-hall, line 93: You still want radius *= Radius_MULTIPLIER.

AbakirH: Yep that makes sense

PR #88, opened 2020-08-03
Final Frontend Changes

ethanimooney:
**TL;DR:** Phew... Lots of small changes, styled search bar, popup, added information div, re styled cards
**LINK:** https://step2020-smallcity.appspot.com/

_Note: The responsiveness of the site is still not perfect, but I think it is good enough for our purposes at the moment. Obviously, if you notice something very noticeable please let me know and I can fix it._ 

**index.html**: Styled initial popup, search bar, loader, and information div. Added centering divs for loader and popup to ensure they're in the exact center of the page. Changed header from header to div. Other various responsiveness updates. Changed zipcode box in header to just a div of text, away from being an input.

**mapScript.js**: Updated to reflect changes in index and script.

**paginationScript.js** - Added a new line apparently...

**script.js**: Changed the way that the popup and loader are displayed, allowing for them to be centered on the screen. Updated script to reflect change of zipcode box in the header to div from input field. Updated the way that results-cards look, image now takes up full left column, moved rating to right side above website button. Added display and hide functions for the information div

**style.css**: Many various responsiveness changes, re-oredering of results-card, added information div and popup styling, re-styled header, etc.

### DIFF
--- a/src/main/webapp/style.css
+++ b/src/main/webapp/style.css
@@ -348,6 +348,18 @@ Main Content Container Styling
   100% { transform: rotate(360deg); }
 }
 
+/* Mini Loader Circle */
+#mini-loader-circle {
+  animation: spin 2s linear infinite;
+  border: 5px solid var(--dark-gray); 
+  border-radius: 50%;
+  border-top: 5px solid var(--main-color); 
+  height: 20px;
+  margin: 0 0 10px 0;
+  opacity: 1;
+  width: 20px;
+}
+
 /*
   Results Content Container Styling
 */
@@ -435,7 +447,14 @@ Main Content Container Styling
 }
 
 .unavailable-website {
+  align-items: center;
   background-color: var(--dark-gray);
+  border: none;
+  border-radius: 30px;
+  cursor: pointer;
+  font-family: 'Raleway', sans-serif;
+  font-size: large;
+  font-weight: bold;
   pointer-events: none;
 }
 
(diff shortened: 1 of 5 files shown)


ethanimooney:
> I think this looks good but there are a few tweaks here and there that I noticed. The most prominent of these comes from the swap of the loader circle for the text box. I don't think this is a bad change, but the loading circle gave the idea that something was processing, the current text box does not. I noticed the line "Finding small buisnesses..." but this is fairly hard to see, and doesn't give the idea that the request is being processed. What about adding an animation to it, in addition to creating more contrast between that line and the white background?

I agree, this was something I thought about doing but was too late in the day for my brain to handle, will add tomorrow. Also will fix a bug with the information doc I noticed later

ethanimooney:
Alright, update:

Fixed the bug I found with the information div being displayed at incorrect times
Merged in the latest master
Changed the zipcode box to look more clickable, added "cursor: pointer;" to all clickable areas
Added the loader circle to the information div
Added ability for user to click off the information div if they don't want to read it, thus displaying the large loader circle

ethanimooney:
Changes are live on the deployed app

src/main/webapp/style.css
```diff
@@ -487,12 +668,43 @@ Main Content Container Styling
...
     }
+
+    #popup-form-container {
+      height: 40%;
+      width: 15%;
+    }
+}
+
+@media screen 
+  and (max-device-height: 900px) {
```
shaquille-hall, line 687: nit: I think we talked about preferring max-[height|width] here, and below.

ethanimooney: Fixed

PR #89, opened 2020-08-04
changed localStorage to sessionStorage

AbakirH:
Summary: Instead of storing the latest search results information forever, we opted to store the results only for the session the user is on the application.


ethanimooney:
I fixed the merge conflict with script.js since I was the one who created it on my end haha

